PR #8, opened 2016-05-13
Moving MParticle-specific messaging logic into a router for reusability

dallasgutauckis:
Core changes:

 - `PushAnalyticsReceiver#onNotification*` methods are now part of an interface (`PushAnalyticsReceiverCallback`)
 - `PushAnalyticsReceiver` now delegates processing of the intent and calling of callbacks into `MPMessagingRouter`
 - `MPMessagingRouter` abides by a contract of only consuming what it should, and lets the caller know if the given `Intent` was consumed

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,117 +2,89 @@
 
 ## Android SDK
 
-Hello! This is the public repo of the mParticle Android SDK. Right now, it contains the "kit" implementations and maintains a dependency on the mParticle Core library. 
-
-We originally built the [mParticle platform](https://www.mparticle.com) to help simplify the approach to SDKs and improve app performance. We've found that having a bunch of SDKs in an app bloats its size and is no-good for your users' battery. So, we built client-side APIs for you to collect all of your data, and a server platform that lets you send your data where you want it to go.
-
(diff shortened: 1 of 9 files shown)


sdk/src/main/java/com/mparticle/messaging/PushAnalyticsReceiverCallback.java
```diff
@@ -0,0 +1,22 @@
+package com.mparticle.messaging;
+
+public interface PushAnalyticsReceiverCallback {
```
dallasgutauckis: This might make more sense to share the `MPMessagingRouter` prefix instead of the Receiver implementation.

samdozor:
hey @dallasgutauckis - first of all, thanks again for the PR. The original intention that we had was for developers optionally subclass `PushAnalyticsReceiverCallback` and override `onNotificationReceived` and/or `onNotificationTapped`. Is there a reason why this pattern doesn't work for your setup?

dallasgutauckis:
I'm assuming you mean subclassing PushAnalyticsReceiver — didn't see the Callback you mentioned when I was looking the other day —  and yeah I understand the suggested implementation. 

Besides preferring not to depend on third party SDKs for our super classes (composition over inheritance), some other things:

 - Calling the super implementation doesn't let us know if the intent provided is handled by MParticle or not
 - Our existing GCM receiver implementation utilizes WakefulBroadcastReceiver, and with this implementation we'll still be sending our own messages

>  I think that accomplishes the same thing, but maybe I'm missing something

It's close, but not quite the same. I would say the main benefit is knowing if the result should be handled by MParticle

PR #22, opened 2018-04-19
Add consumer proguard rule to guard reflective access

asimone0:
[MParticleUtility.getGoogleAdIdInfo()](https://github.com/mParticle/mparticle-android-sdk/blob/master/android-core/src/main/java/com/mparticle/internal/MPUtility.java#L113) attempts to reference com.google.android.gms.ads.identifier.AdvertisingIdClient using reflection

Guarding the package with a proguard rule increases the likelihood of success for the library code

### DIFF
--- a/android-core/consumer-proguard.pro
+++ b/android-core/consumer-proguard.pro
@@ -1,3 +1,4 @@
 #these are the rules that are packaged with the Core SDK .aar to be used in consuming apps.
 -keep class com.mparticle.** { *; }
 -dontwarn com.mparticle.**
+-keep class com.google.android.gms.ads.identifier.** { *; }


samdozor:
Hi @asimone0 ! Thanks for the PR. I'm curious...did you run into this being an issue? Per Google's docs[1], they include their own proguard rules as well so strictly speaking this shouldn't be necessary:

> Note: ProGuard directives are included in the Play services client libraries to preserve the required classes. The Android Plugin for Gradle automatically appends ProGuard configuration files in an AAR (Android ARchive) package and appends that package to your ProGuard configuration. During project creation, Android Studio automatically creates the ProGuard configuration files and build.gradle properties for ProGuard use. To use ProGuard with Android Studio, you must enable the ProGuard setting in your build.gradle buildTypes. For more information, see the ProGuard guide.

[1]https://developers.google.com/android/guides/setup#add_google_play_services_to_your_project

asimone0:
@samdozor 
I have seen the issue first hand and reported it to a mParticle technical engineer last fall via slack.

The general scenario was as follows:
The application **_using_** the mParticle library is able to obtain and use the advertiser ID without issue.
However, the mParticle library itself is not able to obtain it.

Before the rule:

`D/mParticle: Failed to collect Google Play Advertising ID, be sure to add Google Play services or com.google.android.gms:play-services-ads to your app's dependencies.`

Adding the proguard rule in the *application* proguard config was necessary in order for the mParticle library code to obtain the advertiser id.

After the rule:
`D/mParticle: Successfully collected Google Play Advertising ID.`

My hypothesis is that the reflective approach in MParticleUtility allows proguard to conclude the class itself is not needed and remove it.

asimone0:
Likely the case.
Thanks for your time

PR #28, opened 2018-10-23
Update Proguard config

rossbacher:
## Summary

It's not a good idea to keep `com.mparticle.** { *; }` because even though the SDK is already proguarded there are still parts of it that might not be used by an app implementing it and those need to be removed.

Added rule to keep the Advertisment Id code as it is accessed via reflection from mParticle SDK and needs to be kept in order access it.

## Testing Plan

Still needs to be tested, this is just a suggestion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ public class MyApplication extends Application {
 
 #### Proguard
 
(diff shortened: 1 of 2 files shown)


samdozor:
Hey @rossbacher thanks for the PR! There are a few reasons why we do this today
1. We use reflection to detect the presence of kits. Without it, customers would have to add each kit to their exclusion rules.
2. As you noted the library is already proguarded. This hopefully means that the advantage you get of proguarding it again will be minimal. Are you seeing a big improvement in method count with this change?
3. The SDK is unit and integration tested with this configuration, but we can't test customers' custom proguard rules. In the past we've seen Proguard to lead to some very tricky bugs and crashes.

Unfortunately it doesn't look like the android gradle plugin provides a way to override an AARs consumer-proguard file (short of you cloning and building the SDK yourself, which is an option). We can definitely consider this change though - what sort of improvement are you seeing on your end with this in place?

samdozor:
Closing this due to inactivity.

PR #72, opened 2021-11-02
ci: update PR template and fix release job sync step

mchuangatmp:
## Summary
- Fix release job last step and update PR template doc

## Testing Plan
- Release job should run successfully

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,8 @@
+## Instructions
+1. PR target branch should be against development
+2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
+3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml
+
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,8 @@
+## Instructions
+1. PR target branch should be against development
+2. PR title name should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-title-check.yml
+3. PR branch prefix should follow this format: https://github.com/mParticle/mparticle-workflows/blob/main/.github/workflows/pr-branch-check-name.yml
```
willpassidomo, line 4: we should link to the same file that we are actually running for the job. this link is to `main`, but should be `stable`

I think it's just as simple as: 
https://github.com/mParticle/mparticle-workflows/blob/stable/.github/workflows/pr-title-check.yml
and 
https://github.com/mParticle/mparticle-workflows/blob/stable/.github/workflows/pr-branch-check-name.yml

mchuangatmp: good suggestion...I updated PR with change

willpassidomo (approved):

PR #80, opened 2021-12-14
ci: migrate to sonarcloud

mchuangatmp:
## Summary
- Switch from SonarQube to SonarCloud
- Code Cleanup
- Change `localbroadcastmanager` from `implementation` to `api`

## Testing Plan


willpassidomo:
@mchuangatmp is this on the right base branch? If not, this should be broken up into at least 3 PRs (soarcloud, dependency change, formatting) but I haven't read through the whole thing

PR #81, opened 2021-12-15
ci: add in sonarcloud check

mchuangatmp:
## Summary
- Switch from broken SonarQube to SonarCloud

## Testing Plan


mchuangatmp:
I will reopen this once i can get tests to pass more consistently (so actions don't keep running)

PR #82, opened 2021-12-15
fix: change localbroadcastmanager back to compile in pom.xml

mchuangatmp:
## Summary
- Change `localbroadcastmanager` from `implementation` to `api` to output `compile` instead of `runtime` in `pom.xml` file.  Our previous AGP 7 upgrade changed some behavior.

## Testing Plan
- normal regression tests should pass


mchuangatmp:
@willpassidomo looks like we should merge in the sonarcloud PR first, since tests fail for this one

willpassidomo (approved):

mchuangatmp:
not needed anymore, will be changed in a future PR

PR #83, opened 2021-12-17
ci: update sonarcloud and dependencies

mchuangatmp:
## Summary
- Switch from broken SonarQube to SonarCloud

## Testing Plan


mchuangatmp:
target branch is caching so I will recreate same PR to fix

PR #85, opened 2021-12-21
feat: Add configMaxAge to MParticleOptions

willpassidomo:
(moved from internal repo)

## Summary
Add public API for setting a max config staleness parameter. If the config is older than staleness limit, we will delete it instead of loading it from memory while we wait for the new config

### Usage

```
val options = MParticleOptions.builder(context)
   .maxConfigAge(1000 * 60 * 60 * 24 * 7)                   <- set a max age of 1 week
   .build()
```
note: **maxConfigAge value is ~DAYS~ SECONDS**

### API Behavior

- **not setting** maxConfigAge will default it to `null` which means it will **not** take effect and the SDK will behave exactly how it currently does
- setting it to **a negative Integer** will trigger a warning-level log message then proceed as if it was `null` (same as above)
- setting it to `0` will result in the stored config being deleted every time the SDK starts up aka Kits et al we will behave like it's a fresh install every time
~- setting it to **an Integer greater than 100** will trigger a warning-level log message reminding the client that the time unit is days, but still proceed with the value~
- setting it to **a reasonable (<100) Integer** will set that value as the config max-age in ~days~ seconds

### Internal Behavior
- we will check config age exactly once per app start in the ConfigManager constructor
> although it is possible the ConfigManager may get called at other points in our code, no where else will it pass in a configMaxAge, so in those cases it would default to `null`/do nothing
- staleness is determined by `("config timestamp" + "Milliseconds in a day" * configMaxAge) < System.currentTimeMillis`so there is no rounding or anything, the config will be invalid
- when a config is "deleted" for staleness reasons, we **will also delete the stored ETag and stored If-Modified values**.

## Testing Plan
- unit tests for `MParticleOptions` and `ConfigManager`
- integration tests for `maxConfigAge == 0` and "maxConfigAge not set" where we 1) start the SDK 2) wait for config to load 3) stop SDK 4) restart SDK 5) check stored config before the new one is fetched
- integration test for maxConfig == (big #) and make sure it isn't deleted before it should 
- new listener in ComfigManager that will callback when a config is "loaded" and also indicate whether it was a "reload" or a  new config...needed this so we had something to actually wait on when refreshing the config in test

~//TODO~
~- mock time and add an integration test for maxConfig == (some number), change the internal time and verify it gets properly deleted~

## Master Issue
Closes https://mparticle.tpondemand.com/RestUI/Board.aspx#page=board/4842431717611364328&appConfig=eyJhY2lkIjoiMjM4RjY2OUU2RkYxNzFCMTVCMUU5Q0JCRDlFQzQyRTAifQ==&boardPopup=bug/71594/silent

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
@@ -31,12 +31,15 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 
 public class ConfigManager {
     public static final String CONFIG_JSON = "json";
+    public static final String CONFIG_JSON_TIMESTAMP = "json_timestamp";
     private static final String KEY_TRIGGER_ITEMS = "tri";
     private static final String KEY_MESSAGE_MATCHES = "mm";
     private static final String KEY_TRIGGER_ITEM_HASHES = "evts";
@@ -96,10 +99,12 @@ public class ConfigManager {
     private JSONObject mCurrentCookies;
     private String mDataplanId;
     private Integer mDataplanVersion;
+    private Long mMaxConfigAge;
     public static final int DEFAULT_CONNECTION_TIMEOUT_SECONDS = 30;
     public static final int MINIMUM_CONNECTION_TIMEOUT_SECONDS = 1;
     public static final int DEFAULT_SESSION_TIMEOUT_SECONDS = 60;
     public static final int DEFAULT_UPLOAD_INTERVAL = 600;
+    private List<ConfigLoadedListener> configUpdatedListeners = new ArrayList<ConfigLoadedListener>();
 
     private ConfigManager() {
         super();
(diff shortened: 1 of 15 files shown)


mchuangatmp (approved):

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -212,25 +224,92 @@ static void deleteConfigManager(Context context) {
...
+            return sPreferences.getLong(CONFIG_JSON_TIMESTAMP, 0);
+        } else {
+            return null;
+        }
+    }
+
+    void saveConfigJson(JSONObject json, String etag, String lastModified) throws JSONException {
         if (json != null) {
-            sPreferences.edit().putString(CONFIG_JSON, json.toString()).apply();
+            Logger.info("Updating config to:\n" + json.toString());
```
samdozor, line 270: can you change this to debug log level (or whichever is the most verbose)? it may contain api keys

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -212,25 +224,92 @@ static void deleteConfigManager(Context context) {
...
+            sPreferences.edit()
+                    .putLong(CONFIG_JSON_TIMESTAMP, System.currentTimeMillis())
```
samdozor, line 272: noting that we could use the timestamp that the server returns with if we wanted, but I actually think using the device timestamp is better - in case the device timestamp is inaccurate at least the delta/comparison will be correct.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -212,25 +224,92 @@ static void deleteConfigManager(Context context) {
...
+    @Nullable
+    Long getConfigTimestamp() {
+        if (sPreferences.contains(CONFIG_JSON_TIMESTAMP)) {
+            return sPreferences.getLong(CONFIG_JSON_TIMESTAMP, 0);
+        } else {
+            return null;
+        }
+    }
+
+    void saveConfigJson(JSONObject json, String etag, String lastModified) throws JSONException {
```
samdozor, line 268: I like moving the responsibility to persist these here. That said - there is a slight behavior change in that we used to check if they were null. This might be better - whereby if they are null we will pass null into putString below - but then we will also get null back out whereas behavior maybe we wouldn't have? Is there any concern when these are null?

samdozor (dismissed):

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
willpassidomo: agreed, I think it's pretty in line with the overall theme of this change; create some client-side/server-independent checks on config state

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
willpassidomo: thats a good call-out, it is a change in existing behavior. When I came across the previous logic, it looked like a bug to me, although not a particularly bad one. If we fetch "config1" which returns with "etag1", then the next time around we fetch "config2" which returns _without_ an etag, on the next request we will be using the stale "etag1" in our request header which may lead the server to believe our config is stale-er than it actually is. Net-net, the way I read it, the previous behavior may lead to us fetching more configs than we need too (the extra should have been 304s).

That said, I don't see any concern over this change. Sans any comments or notes on some unexpected behavior we were accounting for when we originally implemented this, this new version seems more "correct"

PR #89, opened 2022-01-05
feat: Android id disabled default true

willpassidomo:
#### Breaking Change!!

Change the default value for `MParticleOptions.isAndroidIdDisabled()` and `MParticle.isAndroidIdDisabled()` to `true`. Previously the default was false.

The net result will be that any client (including existing clients) who do not explicitly call `MParticleOptions.Builder.androidIdDisabled(Boolean)` will see the SDK behavior change. Previously, for these clients, we included Android ID as a `known_identities` entry in Identity Requests as well as a Device Attribute field in Batch Messages, but with this change we will only do so if they set `androidIdDisabled(true)` when building `MParticleOptions`

#### API Change

We have deprecated `MParticleOptions#isAndroIdDisabled()` and `MParticle.isAndroidIdDisabled()`. These methods were replaced by`MParticleOptions#isAndroidIdEnabled()` and `MParticle.isAndroidIdEnabled()`, respectively 

## Testing Plan
updated existing tests, added unit tests for `MParticleOptions` to check that default is correct and setter still works, added integration tests to make sure the setting properly propegated

## Master Issue
Closes https://go.mparticle.com/work/77592

### DIFF
--- a/android-core/src/androidTest/java/com/mparticle/MParticleOptionsTest.java
+++ b/android-core/src/androidTest/java/com/mparticle/MParticleOptionsTest.java
@@ -155,17 +155,30 @@ public class MParticleOptionsTest extends BaseAbstractTest {
     @Test
     public void testAndroidIdDisabled() throws Exception {
         //test defaults
+        assertFalse(MParticle.isAndroidIdEnabled());
         assertTrue(MParticle.isAndroidIdDisabled());
         MParticle.setInstance(null);
         startMParticle(MParticleOptions.builder(mContext));
+        assertFalse(MParticle.isAndroidIdEnabled());
         assertTrue(MParticle.isAndroidIdDisabled());
 
-        //test androidIdDsabled == true
+        //test androidIdDisabled == true
         MParticle.setInstance(null);
         startMParticle(
                 MParticleOptions.builder(mContext)
                         .androidIdDisabled(true)
         );
+        assertFalse(MParticle.isAndroidIdEnabled());
+        assertTrue(MParticle.isAndroidIdDisabled());
+        MParticle.setInstance(null);
+
+        //test androidIdEnabled == false
+        MParticle.setInstance(null);
+        startMParticle(
+                MParticleOptions.builder(mContext)
+                        .androidIdEnabled(false)
+        );
+        assertFalse(MParticle.isAndroidIdEnabled());
         assertTrue(MParticle.isAndroidIdDisabled());
         MParticle.setInstance(null);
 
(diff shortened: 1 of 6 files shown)


android-core/src/androidTest/java/com/mparticle/MParticleOptionsTest.java
```diff
@@ -425,4 +437,9 @@ public void testNetworkOptions() {
         assertTrue(com.mparticle.networking.AccessUtils.equals(options.getNetworkOptions(), com.mparticle.networking.AccessUtils.getDefaultNetworkOptions()));
     }
 
+    @Test
+    public void testAndroidId() {
+
```
mchuangatmp, line 442: @willpassidomo is this test meant to be empty like this?

android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -446,17 +457,33 @@ public Builder devicePerformanceMetricsDisabled(boolean disabled) {
         }
 
         /**
-         * By default, the SDK will collect <a href="http://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID">Android Id</a> for the purpose
-         * of anonymous analytics. If you're not using an mParticle integration that consumes Android ID, the value will be sent to the mParticle
-         * servers and then immediately discarded. Use this API if you would like to additionally disable it from being collected entirely.
+         * By default, the SDK will NOT collect <a href="http://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID">Android Id</a> for the purpose
+         * of anonymous analytics. If you're not using an mParticle integration that consumes Android ID and you would like to collect it, use this API to enable collection. The value will be sent to the mParticle
```
samdozor, line 461: can you add something like this too

```
 * @deprecated
 * This method has been replaced as the behavior has been inverted - Android ID collection is now disabled by default. 
 * <p> Use {@link androidIdEnabled(boolean)} instead.
```

android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -446,17 +457,33 @@ public Builder devicePerformanceMetricsDisabled(boolean disabled) {
...
+        @Deprecated
         public Builder androidIdDisabled(boolean disabled) {
-            this.androidIdDisabled = disabled;
+            this.androidIdEnabled = !disabled;
+            return this;
+        }
+
+        /**
+         * By default, the SDK will NOT collect <a href="http://developer.android.com/reference/android/provider/Settings.Secure.html#ANDROID_ID">Android Id</a> for the purpose
+         * of anonymous analytics. If you're not using an mParticle integration that consumes Android ID and you would like to collect it, use this API to enable collection. The value will be sent to the mParticle
```
samdozor, line 477: this is tricky - it's not actually true that we'd discard it. i'd just kill this comment

samdozor (dismissed):

android-core/src/main/java/com/mparticle/MParticleOptions.java
willpassidomo: cool, I thought that sounded a little off. fixed now

samdozor (approved):
Looks good!

Just a reminder that:
- this should be a new minor version
- let me know when you're getting ready to release and I'll be ready to add a very loud changelog message.

PR #90, opened 2022-01-05
ci: create first version of daily android health check cron job

mchuangatmp:
## Summary
- Add a daily android health check cron job (for releases)

## Testing Plan
- Running cron from workflow dispatch should pass

## Master Issue
- Closes https://go.mparticle.com/work/77650

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,0 +1,154 @@
+name: "Daily Cron"
+
+on:
+  # Run daily at midnight
+  schedule:
+    - cron: "0 0 * * *"
+
+  # Allow workflow to be manually run from the GitHub UI
+  workflow_dispatch:
+
+jobs:
+
+  instrumented-tests:
+    name: "Instrumented Tests"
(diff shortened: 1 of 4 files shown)


.github/workflows/daily.yml
```diff
@@ -0,0 +1,129 @@
...
+      - name: "Run Unit Tests"
+        run: ./gradlew test
+      - name: "Android Test Report"
+        uses: asadmansr/android-test-report-action@v1.2.0
+        if: ${{ always() }}
+
+  update-kits:
+    name: "Update Kits"
+    needs: [instrumented-tests, unit-tests]
+    runs-on: macos-11
```
willpassidomo, line 64: I forgot why we originally chose it, but is it possible to run this on ubuntu?

mchuangatmp: it was missing a dependency when we last tried (just faster to use macOS since we know it works)...we can try again, but let's make it part of another PR if it works

willpassidomo (approved):
Awesome!

PR #96, opened 2022-01-12
chore: Add GA4 submodule

rmi22186:
## Instructions

## Summary
This PR adds GA4 as a submodule so that it can be auto released during the core SDK release job.

Note that this is in draft for now because I am not able to add the submodule itself until it is public, and it is currently set as `internal` because we need to release web/server/ios/android all at the same time.

## Master Issue
Closes https://go.mparticle.com/work/78219


rmi22186:
Marking this as ready for review so you guys can look at it.  However, it should NOT be merged until GA4 is first public, so that we can add another commit here to add the submodule.

mchuangatmp:
> Marking this as ready for review so you guys can look at it. However, it should NOT be merged until GA4 is first public, so that we can add another commit here to add the submodule.

LGTM but can you rebase the branch so it has latest commits?  or do you need Will or me to do it?

rmi22186:
done @mchuangatmp

rmi22186:
Next steps:  we will wait to add the GA4 submodule when that is public. I'll add a commit here, then put this back up for approval.

willpassidomo (commented):
Looks good! Approved informally, just need to wait on the GA4 submodule before merging

mchuangatmp (approved):

PR #98, opened 2022-01-20
fix: update java docs syntax after gradle upgrade

mchuangatmp:
## Summary
- update java docs syntax after gradle upgrade

## Testing Plan
- CI tests should pass and javadocs should also generate

## Master Issue
Closes https://go.mparticle.com/work/78794


willpassidomo (approved):
🙌

PR #100, opened 2022-02-18
feat: remove deprecated methods, AppConfig

willpassidomo:
## Summary
This removes some unused functionality left over from when we gave the option of configuring the SDK via XML resources. Attempting to read these values is fairly slow, and this code was a big culprit in causing our slow startup issues. 

After removing the code, we are able to effectively merge `AppConfig` and `ConfigManager` which reduces our complexity.

In addition to removing the resource fetching, we removed a fallback check for stored credentials in `MParticleOptions`. Primarily used as a mechanism for internal restarts, we would allow `MParticleOptions` to build, without credentials, as long as there were credentials in persistence. Since client implementations are static, we don't believe there is an external use-case for this check. Since the reads were slowing down initialization of this object, we instead opened up a new internal endpoint, `MParticleOptions.Builder#buildForInternalRestart()` which allows us to bypass the credential check for internal restarts

## Testing Plan
existing tests cover most of what we need. I replaced the MParticleOptions initialization method in. a few spots, but for the most part there are just updates

## Master Issue
Closes https://go.mparticle.com/work/80825


testutils/src/main/java/com/mparticle/networking/MockServer.java
```diff
@@ -419,16 +419,16 @@ private MPUrl getUrl(String identityPath, Long mpid) {
     private MParticleBaseClientImpl getClient() {
         ConfigManager configManager = null;
```
mchuangatmp, line 420: i think we can delete this ConfigManger variable if it's not being used

mchuangatmp (approved):

android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -639,7 +639,7 @@ public void enableLocationTracking(@NonNull String provider, long minTime, long
                     //noinspection MissingPermission
                     locationManager.requestLocationUpdates(provider, minTime, minDistance, mLocationListener);
                     locationTrackingEnabled = true;
-                }catch (SecurityException se) {
+                } catch (SecurityException se) {
```
mchuangatmp, line 642: this should be ignored 

`} catch (SecurityException ignored) { }`

testutils/src/main/java/com/mparticle/networking/MockServer.java
willpassidomo: 👍

android-core/src/main/java/com/mparticle/MParticle.java
willpassidomo: 👍

PR #103, opened 2022-02-22
feat: split configuration into core and kits

willpassidomo:
## Summary
`SharedPreferences` loads pretty slow. Worse than that, it seems to grow linearly with the overall size of the file. The edge cases where we see very slow startup time are generally in cases where the user has a large number of Kit filters or projections, which lead to a very large config (stored in `SharedPreferences`) and result in a slow loading time when we fetch our first pair from the file. Outside of the various Kit configuration properties, the rest of the config has a fairly low ceiling for how big it could possibly get.

Since we don't actually need the kit portion of the configuration at startup, it makes sense to store it outside of `SharedPreferences` so we can load _just_ the data we need at startup. This PR splits the config when it is fetched and continues to persist the "core config" (everything outside of `"eks"`) in `SharedPreferences`, but saves the "kit config" (`"eks"`) in a separate file. Instead of being loaded on SDK initialization, the kit config will not be loaded in `delayedStart()`. Additionally, we reduced the delay for `delayedStart()` from 20 seconds to 1 second (reviewed w/Sam)

Additionally, we added a migration step that will remove any old kit config that may be stored in `SharedPreferences` when the SDK is initialized

## Testing Plan
Added tests coving the migration logic, fixed the flakey `PustRegistrationTest` case

## Master Issue
Closes https://go.mparticle.com/work/80843

### DIFF
--- a/android-core/src/androidTest/java/com/mparticle/PushRegistrationTest.java
+++ b/android-core/src/androidTest/java/com/mparticle/PushRegistrationTest.java
@@ -163,9 +163,17 @@ public class PushRegistrationTest extends BaseCleanStartedEachTest {
                     //For enablePushNotifications() to set the push registration, we need to mimic
                     //the Firebase dependency, and clear the push-fetched flags
                     TestingUtils.setFirebasePresent(true, pushRegistration.instanceId);
-                    ConfigManager.getInstance(mContext).clearPushRegistration();
-
                     MParticle.getInstance().Messaging().enablePushNotifications(pushRegistration.senderId);
+                    //this method setting push is async, so wait for confirmation before continuing
+                    ConfigManager configManager = ConfigManager.getInstance(mContext);
+                    while (!configManager.isPushEnabled()) {
+                        try {
+                            Thread.sleep(10);
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
+                        }
+                    }
+                    TestingUtils.setFirebasePresent(false, null);
                 }
 
                 @Override
@@ -249,7 +257,7 @@ public class PushRegistrationTest extends BaseCleanStartedEachTest {
 
                 @Override
                 public String getName() {
-                    return null;
+                    return "startMParticle(MParticleOptions.builder(mContext).pushRegistration(null, null))";
                 }
             }
     };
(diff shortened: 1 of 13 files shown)


android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -268,53 +289,88 @@ Long getConfigTimestamp() {
...
+        updateCoreConfig(responseJSON, false);
     }
 
-    private synchronized void updateConfig(JSONObject responseJSON, boolean newConfig, String etag, String lastModified) throws JSONException {
+    private synchronized void reloadKitConfig(@Nullable JSONArray kitConfigs) {
+        //only reload if KitManager has not already been loaded (from new config, presumably)
         MParticle instance = MParticle.getInstance();
-        KitFrameworkWrapper kitManager = null;
-        if (instance != null) {
-            kitManager = instance.Internal().getKitManager();
```
willpassidomo, line 311: no need to null-check here, `Internal().getKitManager()` are both non-null

android-core/src/androidTest/java/com/mparticle/PushRegistrationTest.java
```diff
@@ -163,9 +163,17 @@ public void setPushRegistration(PushRegistrationHelper.PushRegistration pushRegi
                     //For enablePushNotifications() to set the push registration, we need to mimic
                     //the Firebase dependency, and clear the push-fetched flags
                     TestingUtils.setFirebasePresent(true, pushRegistration.instanceId);
-                    ConfigManager.getInstance(mContext).clearPushRegistration();
-
                     MParticle.getInstance().Messaging().enablePushNotifications(pushRegistration.senderId);
+                    //this method setting push is async, so wait for confirmation before continuing
+                    ConfigManager configManager = ConfigManager.getInstance(mContext);
+                    while (!configManager.isPushEnabled()) {
```
mchuangatmp, line 169: wouldn't a listener as a callback be better than a while loop?  This is a test though so maybe it doesn't matter as much

mchuangatmp:
@einsteinx2 if you have time, I think you should review some of the proposed architectural code changes in this PR (to optimize startup time)

android-core/src/androidTest/java/com/mparticle/PushRegistrationTest.java
willpassidomo: it would..but it's one of those things that isn't actually valuable to the client as far as I can tell... probably not worth going through the trouble of adding one right now.

PR #114, opened 2022-03-08
fix: add etag to batch upload payload

mchuangatmp:
## Instructions

## Summary
- add etag to batch upload

## Testing Plan
- make sure sdk doesn't break in a sample app with new key (server should ignore for now)

## Master Issue
- Closes https://go.mparticle.com/work/81946


einsteinx2 (approved):
LGTM. Afaik the keys are arbitrary and in fact these "short keys" are then converted to "long keys" on the backend (something we would implement later), and are only used as a way to decrease network data usage, making them especially arbitrary since they're not even the final keys used in the system.

willpassidomo (approved):

PR #122, opened 2022-04-19
fix: add in missing projection fields

mchuangatmp:
## Instructions

## Summary
- Add in missing projection fields

## Testing Plan
- Test a sample app that connects to a workspace that has custom mapping fields.  `af_content_view` should show up in timeline

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-3404


willpassidomo (approved):
good find!

PR #123, opened 2022-04-19
fix: update test kit command in release.yml

mchuangatmp:
## Instructions

## Summary
- update test kit command in release.yml

## Testing Plan
- release job should work

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-3708


willpassidomo (approved):
good fix!

PR #141, opened 2022-05-18
feat: add client side rules

mchuangatmp:
## Instructions

## Summary
- add client side rules

## Testing Plan
- test in sample app editing one event batch

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-3734

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -42,6 +42,7 @@ public class MParticleOptions {
     private Boolean mUnCaughtExceptionLogging = false;
     private MParticle.LogLevel mLogLevel = MParticle.LogLevel.DEBUG;
     private AttributionListener mAttributionListener;
+    private BatchCreationListener batchCreationListener = null;
     private LocationTracking mLocationTracking;
     private PushRegistrationHelper.PushRegistration mPushRegistration;
     private Integer mIdentityConnectionTimeout = ConfigManager.DEFAULT_CONNECTION_TIMEOUT_SECONDS;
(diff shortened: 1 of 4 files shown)


android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -940,4 +956,15 @@ public DataplanOptions build() {
...
+
+    /**
+     Custom handler to modify or block batch data before upload.
+     If set, this will be called when a new batch of data is created. By returning a different value, you can change the batch contents, or by returning 'nil' you can block the batch from being uploaded.
+     Use with care. This feature was initially added to allow the value of existing fields to be modified. If you add new data in a format that the platform is not expecting, it may be dropped or not parsed correctly.
+     Note: Use of this handler will also cause the field 'mb' (modified batch) to appear in the batch so we can distinguish for troubleshooting purposes whether data was changed.
+     Also note: Unlike other callbacks, this block will be called on the SDK queue to prevent batches from being processed out of order. Please avoid excessively blocking in this handler as this will prevent the SDK from doing other tasks.
+     */
+    public interface BatchCreationListener {
+        JSONObject onBatchCreated(JSONObject batch);
```
willpassidomo, line 968: nit: Annotate the return and parameter types, this will make our Kotlin interop a lot smoother :)

android-core/src/main/java/com/mparticle/internal/database/services/MParticleDBManager.java
```diff
@@ -300,7 +309,17 @@ private void createUploads(Map<BatchId, MessageBatch> uploadMessagesByBatchId, M
                 uploadMessage.setIdentities(identities);
                 JSONObject userAttributes = findUserAttributeState(messages, batchId.getMpid());
                 uploadMessage.setUserAttributes(userAttributes);
-                UploadService.insertUpload(db, uploadMessage, configManager.getApiKey());
+
+                JSONObject batch = uploadMessage;
+                if (options.getBatchCreationListener() != null) {
+                    batch = options.getBatchCreationListener().onBatchCreated(batch);
+                }
```
willpassidomo, line 316: we should wrap this in a try/catch so the client code doesn't blow up the rest of this procedure if their code throws an exception

willpassidomo (changes requested):

willpassidomo (approved):
Looks great! FF with integration tests before release

PR #145, opened 2022-05-20
fix: change default batch creation error logic

mchuangatmp:
## Instructions

## Summary
- change default batch creation error logic

## Testing Plan
- batch tests should pass

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4076


willpassidomo (approved):
Looks great, test update makes sense

PR #174, opened 2022-06-06
fix: add in missing sync step for dependabot

mchuangatmp:
## Instructions

## Summary
- add in missing sync step for dependabot

## Testing Plan
- release job should still work

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4219


mchuangatmp:
@willpassidomo i'll just add this change in my next PR in this repo (but before next release)

PR #181, opened 2022-06-12
fix: update dependabot token for pull_request

mchuangatmp:
## Instructions

## Summary
- Use github.com's recommend workaround for `pull_request_target` PWN issue

## Testing Plan
- forked PR tests should still run

## Master Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4055

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,5 +1,7 @@
 name: "Build and Test"
-on: [push, pull_request]
+
+on: [ push, workflow_dispatch, pull_request ]
+
 jobs:
   instrumented-tests:
     name: "Instrumented Tests"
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request.yml
```diff
@@ -9,8 +11,8 @@ jobs:
       - name: "Checkout Branch"
         uses: actions/checkout@v2
         with:
-          repository: ${{github.event.pull_request.head.repo.full_name}}
-          ref: ${{github.head_ref}}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
```
willpassidomo, line 15: nit: we can remove the repository/ref fields since we are no longer using `pull_request_target` :)

.github/workflows/pull-request.yml
```diff
@@ -35,8 +37,8 @@ jobs:
...
```
willpassidomo, line 41: same :)

willpassidomo (approved):
Look good to me! Just a couple nit-picks about removing repository/ref since we don't need them anymore, but those won't affect the functionality :)

PR #186, opened 2022-06-16
fix: clean up dependabot step

mchuangatmp:
## Instructions

## Summary
- Filter automerge should only happen for dependabot user

## Testing Plan
- Dependabot automerge should work

## Ticket Number
- Closes https://go.mparticle.com/work/SQDSDKS-4182


willpassidomo (approved):
Looks good to me!

PR #188, opened 2022-06-17
fix: change incrementUserAttribute method signature to accept Number

jshin-tse:
… in line with iOS implementation

## Instructions

## Summary
- Request to change incrementUserAttribute method signature to accept doubles to be consistent with iOS implementation
- https://github.com/mParticle/mparticle-apple-sdk/blob/main/mParticle-Apple-SDK/Identity/MParticleUser.m#L227

## Testing Plan
- Implicit type conversions should do most of the work, though maybe an overloaded incrementUserAttribute that uses int can be kept for legacy customers that rely on the data type of a user attribute being int?

## Master Issue
- (https://mparticlehelp.zendesk.com/agent/tickets/8336)

### DIFF
--- a/android-core/build.gradle
+++ b/android-core/build.gradle
@@ -90,6 +90,7 @@ android {
         lintConfig file('lint-baseline.xml')
     }
     sourceSets {
+        main.kotlin.srcDirs += 'src/main/kotlin'
         test.java.srcDirs += 'src/test/kotlin'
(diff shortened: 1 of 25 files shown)


willpassidomo:
Hey @jshin-tse, it looks like you still have a few compilation errors. Let me know if you want to walk through this one with me :)

mchuangatmp (approved):

PR #233, opened 2022-09-01
feat: update isLimitAdTracking logic for Android 13

mchuangatmp:
## Instructions

## Summary
- Removed deprecated AAID logic
- Return `null` for GAID when LAT in enabled
- Convert MParticleTest.java to kotlin

## Testing Plan
- Emulator returns 0s when isLimitAdTrackingEnabled = true
- Emulator returns gaid when isLimitAdTrackingEnabled = false
- Regression tests should pass (except crossplatforms since AGP hasn't been bumped up there)

## Reference Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4524

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
@@ -54,7 +54,6 @@ public class ConfigManager {
     public static final String KEY_EMBEDDED_KITS = "eks";
     static final String KEY_UPLOAD_INTERVAL = "uitl";
     static final String KEY_SESSION_TIMEOUT = "stl";
-    public static final String KEY_AAID_LAT = "rdlat";
     public static final String VALUE_APP_DEFINED = "appdefined";
     public static final String VALUE_CUE_CATCH = "forcecatch";
     public static final String PREFERENCES_FILE = "mp_preferences";
@@ -89,7 +88,6 @@ public class ConfigManager {
 
     private boolean mSendOoEvents;
     private JSONObject mProviderPersistence;
-    private boolean mRestrictAAIDfromLAT = true;
     private int mRampValue = -1;
     private int mUserBucket = -1;
 
@@ -422,7 +420,6 @@ public class ConfigManager {
             mInfluenceOpenTimeout = 30 * 60 * 1000;
         }
 
-        mRestrictAAIDfromLAT = responseJSON.optBoolean(KEY_AAID_LAT, true);
         mIncludeSessionHistory = responseJSON.optBoolean(KEY_INCLUDE_SESSION_HISTORY, true);
         if (responseJSON.has(KEY_DEVICE_PERFORMANCE_METRICS_DISABLED)) {
             MessageManager.devicePerformanceMetricsDisabled = responseJSON.optBoolean(KEY_DEVICE_PERFORMANCE_METRICS_DISABLED, false);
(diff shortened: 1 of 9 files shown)


android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1213,7 +1198,7 @@ public void setIdentityApiContext(String context) {
     public String getPreviousAdId() {
         MPUtility.AdIdInfo adInfo = MPUtility.getAdIdInfo(mContext);
         String currentAdId = null;
-        if (adInfo != null) {
+        if (adInfo != null && !adInfo.isLimitAdTrackingEnabled) {
```
samdozor, line 1201: so reading this code - this is what could happen:
1. user has limit ad tracking (LAT) false, and we persist the Ad ID (via `setPreviousAdId` below)
2. they then disable limit ad tracking
3. this method is called - and we skip over this block with your change - so it will return the previous Ad ID, which we should no longer be using 

So that makes me nervous. Do we need these methods at all? Why are we storing the ad ID? If these methods are used to understand if the Ad ID changed, then I would think you'd want this method to return all zeroes or null when LAT is true, so that whatever caller knows that the Ad ID changed from something to null.

Are we triggering a modify call to identity based on these methods? If we are, i'd expect that modify call to remove the Ad ID from the user.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1222,7 +1207,7 @@ public String getPreviousAdId() {
     public void setPreviousAdId() {
         MPUtility.AdIdInfo adInfo = MPUtility.getAdIdInfo(mContext);
         String currentAdId = null;
-        if (adInfo != null) {
+        if (adInfo != null && !adInfo.isLimitAdTrackingEnabled) {
```
samdozor, line 1210: similar situation to above - if the ad id is now null - shouldn't be either be removing the PREVIOUS_ANDROID_ID, setting it to null, all zeroes, or something other than keeping the old one?

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.java
```diff
@@ -256,7 +256,7 @@ public void updateDeviceInfo(Context context, JSONObject deviceInfo) {
                 //check instance nullability here and decline to act if it is not available. Don't want to have the case where we are overriding isLimiAdTrackingEnabled
                 //just because there was a timing issue with the singleton
                 if (instance !=  null) {
-                    if (adIdInfo.isLimitAdTrackingEnabled && instance.Internal().getConfigManager().getRestrictAAIDBasedOnLAT()) {
+                    if (adIdInfo.isLimitAdTrackingEnabled) {
```
samdozor, line 259: the message below is a little wrong now with the latest android OS, Ad ID is not "available" - the library is available, but ad tracking is disabled

samdozor (changes requested):

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.java
mchuangatmp: i changed the copy

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
mchuangatmp: i removed `PREVIOUS_ANDROID_ID` shared pref when LAT is on

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
mchuangatmp: @samdozor in `AppStateManager.java` updated the logic to allow it to send at least one modify call if currentGoogleAdId is `null`

samdozor (approved):
looks good 👍

PR #247, opened 2022-10-06
ci: Update to kotlin 1.7.20

mchuangatmp:
## Instructions

## Summary
- Update to Kotlin 1.7.20
- Add in 3 necessary kit changes
- Add in new dependabot updates

## Testing Plan
- Regression tests should all pass (except crossplatforms since AGP 7.3.0 is not updated in that repo yet)

## Reference Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4616


einsteinx2 (approved):
LGTM

PR #258, opened 2022-10-17
ci: Update github actions that are deprecated

mchuangatmp:
## Instructions

## Summary
- update github actions that are deprecated (set-output warnings are from another repo)

## Testing Plan
- regression suite should pass

## Reference Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4621


BrandonStalnaker (approved):
LGTM

BrandonStalnaker:
Not sure if the SonarCloud Code Analysis is running right but everything else seems good.

PR #287, opened 2022-11-29
fix: Added null check for user attribute value.

FelipeCoro:
## Summary
- Added null check for userAttribute value in FilteredMParticleUser.java onUserAttributesReceived().

## Testing Plan
- All CI test should pass.

## Reference Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4875


markvdouw (commented):
LGTM

markvdouw:
Shound we add/modify unit tests?

PR #288, opened 2022-11-29
fix: Error handle null user attribute value

FelipeCoro:
## Summary
- Added null check for userAttribute value in FilteredMParticleUser.java onUserAttributesReceived(). 

## Testing Plan
- All CI test should pass.

## Reference Issue
- Closes https://go.mparticle.com/work/SQDSDKS-4875


markvdouw (approved):
LGTM but shouldnt we consider this change into unit tests?

PR #295, opened 2022-12-16
feat: Adding WrapperSdk, WrapperSdkVersion - adding this values to MParticleOptions and MParticle

markvdouw:
Adding wrapperSdk in MParticleOptions and MParticle

## Instructions

## Summary
- Adding WrapperSdk enum as definition. Creating abstraction WrapperSdkVersion which contains the WrapperSdk and version.

Add WrapperSdkVersion in MParticle. The default value is NONE with version null, after using the setter to change it to a valid sdkWrapper with a non-null and non-empty version; cant be set again in the same instance (this is a Singleton)
Added tests for different scenarios.

## Testing Plan
- Added unit tests and everything must pass

## Reference Issue
- Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-4757

[SQDSDKS-4757]: https://mparticle-eng.atlassian.net/browse/SQDSDKS-4757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
+++ b/android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
@@ -25,7 +25,6 @@ import org.junit.Test
 import java.io.File
 import java.util.Arrays
 import java.util.concurrent.CountDownLatch
-import kotlin.collections.HashMap
 
 class MParticleTest : BaseCleanStartedEachTest() {
     private val configResponse =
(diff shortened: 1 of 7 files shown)


android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
```diff
@@ -302,6 +305,46 @@ class MParticleTest : BaseCleanStartedEachTest() {
         }
     }
 
+    @Test
+    fun testWrapperSdkVersionWithoutBeingSet() {
+        val instance = MParticle.getInstance()!!
```
mchuangatmp, line 310: is there a better way to do this instead of having so many `!!` in your tests?  or maybe we clean it up afterwards in a future PR?

markvdouw: The instance is nullable by definition. For sure in the code is a horrible practice to force unwrap nullables, in this case the test class on the @Before is making sure that the instance is created, so in this particular situation using the !! is checking only relying on a pre-requisite.
If you want I can change it, using a  ?.let block - but going within the nullable situation should never happen and there it did, I would prefer the test to fail throwing an exception.

markvdouw:
@einsteinx2 What's the Message key abbreviation, and value formatting when uploading the wrapper info?

mchuangatmp (approved):

markvdouw:
@einsteinx2 Can you give a quick look since the impl. changed? Thanks

PR #317, opened 2023-02-03
fix: Removing flaky tests

markvdouw:
Commenting out, removing flaky and inconsistent test difficult to maintain that doesnt add value.

## Instructions

 ## Summary
 - Commenting out, removing flaky and inconsistent test difficult to maintain that doesnt add value.

 ## Testing Plan
 - Unit and Instrumented test should always pass

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5060


samdozor (changes requested):
As we chatted...

(1) looks like there's still a flakey test failing
(2) make sure you pull in the latest from main branch of the core, which bumps ktlint. or we can ignore that failure knowing that's the reason

samdozor (approved):

PR #339, opened 2023-03-01
feat: SDSDKS-5109 Modernization Interfaces

markvdouw:
## Instructions

 ## Summary
 - This introduces the new concept of interfaces, components and an example of a possible interaction between them.
 - Identity, EventLogging, Uploading were created as single responsibility abstractions/implementations and a mediator used to glue the different components.
 - Please do not merge, this is intended to be reviewed and get feedback. Its more likely subject to change.

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5109


einsteinx2 (commented):
We should set up a time to have another squad call tomorrow before you leave for PTO to go over this live and get feedback before closing this out, but overall looks great!

I'll give it a deeper dive later as well, but this is more or less what you presented before so I'm already on board with the overall direction (and so was the squad).

markvdouw:
I made some changes outline the architecture... strategies to resolve different events to log or upload different type of data types, a mediator with actually glues all of the components and is aligned with the pattern itself, extracted the kit manager to be a delegator, kits to work all the same and mpKit as a concrete implementation with our own needs.
This structure is based on components which can be retrieved using the MParticle facade, so that one is aligned with the OCP being able to extend behavior by adding components.
Each components tries to have single responsibility.

Interfaces are in most cases scoped, having an internal and external interface. The internal interfaces extends the external ones, so that adds on top of the public api, but the abstraction layer is clear enough both for internal and external usage.

The kit abstraction MParticleKit, instead of being an interface was modified to be an abstract class so that an empty default implementation is provided and kits are not force or implement empty functions.
If there are any questions, please let me know!

android-core/src/main/java/com/mparticle/modernization/event_logging/MParticleEventLogging.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.mparticle.BaseEvent
+import com.mparticle.modernization.MParticleComponent
+import java.math.BigDecimal
+
+interface MParticleEventLogging : MParticleComponent {
+    //All common, commerce, screen, pushRegistration, notification, notificationOpened and NetworkPerformance events should be logged with the same function
+    fun logEvent(event: BaseEvent)
+    fun leaveBreadcrumb(breadcrumb: String)
+    fun logError(message: String, params: Map<String, String>? = null)
+    fun logLtvIncrease(valueIncreased: BigDecimal, eventName: String? = null, params: Map<String, String>? = null)
```
jasonmgeorge: Lifetime value increase appears to be a special case of a commerce event. Any chance this could be collapsed into logEvent()?

rmi22186:
As a non-kotlin engineer, it's a bit hard for me to see how this would look from a public API perspective. It would be great to get examples of how a developer implements this vs how it is currently.

android-core/src/main/java/com/mparticle/modernization/MParticleCallback.kt
```diff
@@ -0,0 +1,16 @@
+package com.mparticle.modernization
+
+abstract class MParticleCallback<S, E> {
```
rmi22186, line 3: Does this introduce new behavior for a callback to make callbacks across the SDK more consistent?  Or is it just turning it into an abstract class to make it easier to grok?

android-core/src/main/java/com/mparticle/modernization/core/MParticleMediator.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.mparticle.modernization.uploading.MParticleDataUploader
+import com.mparticle.modernization.uploading.MParticleDataUploaderImpl
+import com.mparticle.modernization.uploading.MParticleUploadingStrategy
+import kotlinx.coroutines.CloseableCoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.asCoroutineDispatcher
+import java.util.concurrent.Executors
+
+class MParticleMediator {
```
alexs-mparticle, line 21: Can you give us a brief overview of what is the purpose of the MParticleMediator? Is it analogous to the mp instance manager of the web sdk, in that it collects all of the instances of the mParticle SDK?

rmi22186: Perhaps adding a comment block above each class would be the way to explain each new class's purpose.

android-core/src/main/java/com/mparticle/modernization/core/MParticleComponent.kt
```diff
@@ -0,0 +1,3 @@
+package com.mparticle.modernization.core
+
+interface MParticleComponent
```
alexs-mparticle, line 3: Will there be common methods or properties that will be shared between components, or is this simply defining an abstract concept of a `Component`?

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
```diff
@@ -0,0 +1,94 @@
...
+     * @param callback
+     */
+    fun logout(@NotNull callback: IdentityCallback)
+
+    /**
+     * Returns the user by id or the current user if [mpId] is null
+     * @param mpId if null returns the current user, if not the user by id
+     * @param callback
+     */
+    fun getUser(@Nullable mpId: Long?, @NotNull callback: IdentityCallback)
```
alexs-mparticle, line 30: in the Web SDK, we have the concept of `getCurrentUser` that I don't believe exists in Native. Is this something we should move away from on the Web SDK to be consistent, or Native also support this?

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
```diff
@@ -0,0 +1,94 @@
...
+    suspend fun login(): IdentityApiResult
+
+    /**
+     * Logout
+     * @return identityApiResult
+     */
+    suspend fun logout(): IdentityApiResult
+}
+
+internal interface InternalIdentity : MParticleIdentity {
```
rmi22186, line 64: I'm a bit confused as to what `InternalIdentity` is.  Is there a reason the `MParticleIdentity` interface above does not include modify and identify, but the `InternalIdentity` interface does?

android-core/src/main/java/com/mparticle/modernization/identity/example/MParticleIdentityImpl.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import com.mparticle.identity.IdentityApiRequest
+import com.mparticle.identity.IdentityApiResult
+import com.mparticle.identity.MParticleUser
+import com.mparticle.modernization.MParticleCallback
+import com.mparticle.modernization.core.MParticleMediator
+import com.mparticle.modernization.identity.IdentityCallback
+import com.mparticle.modernization.identity.InternalIdentity
+
+internal class MParticleIdentityImpl(private val mediator: MParticleMediator) :
```
rmi22186, line 11: Is this introducing new behavior where Identity can be blocking?  Or is this the plan for future state?

rmi22186 (approved):

rmi22186:
My approval is signifying that I have approved.

android-core/src/main/java/com/mparticle/modernization/kit/MParticleKit.kt
```diff
@@ -0,0 +1,33 @@
+package com.mparticle.modernization.kit
+
+import com.mparticle.BaseEvent
+import com.mparticle.modernization.eventlogging.MParticleEventLogging
+import java.math.BigDecimal
+
+abstract class MParticleKit : MParticleEventLogging {
```
rmi22186, line 7: Is there an mParticle Kit equivalent that includes the other non-event logging methods?  (onLogin, onSetUserAttribute, etc)

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
einsteinx2: This method as well as the one on line 43 return the current user if null is passed for `mpId`. So it has the functionality, just not as a dedicated method. Whether it would be better to have a dedicated method or to use this way of doing it can be up for debate, but I think we should standardize on one solution across platforms.

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
einsteinx2: So in general, the concept of having internal extensions to public interfaces is to limit what we make public while still having the functionality for internal use. In this particular case, I'm not sure why these are internal though...

android-core/src/main/java/com/mparticle/modernization/identity/example/MParticleIdentityImpl.kt
einsteinx2: It looks like these are blocking but can only be used within a coroutine which are actually async. So I think this is just another way of doing async code. Mark can provide more context on the specifics of coroutine usage, but I believe these are still meant to ultimately be used asynchronously.

android-core/src/main/java/com/mparticle/modernization/kit/MParticleKit.kt
einsteinx2: I think this is probably just missing the addition of the other interfaces and that actually it should be implementing multiple interfaces, not only the event logging one.

android-core/build.gradle
```diff
@@ -135,6 +135,8 @@ dependencies {
     api 'androidx.annotation:annotation:[1.0.0,)'
     compileOnly 'androidx.core:core:[1.3.2, )'
 
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
```
einsteinx2, line 138: This seems to be breaking cross platform testing CI action.

android-core/src/main/java/com/mparticle/modernization/data/MParticleDataRepository.kt
```diff
@@ -0,0 +1,10 @@
+package com.mparticle.modernization.data
+
+import com.mparticle.internal.messages.BaseMPMessage
+
+class MParticleDataRepository {
```
alexs-mparticle, line 5: Can you explain what the purpose of the Data Repository is?

android-core/src/main/java/com/mparticle/modernization/event_logging/MParticleEventLogging.kt
markvdouw: We could create the abstraction of LtvEvent extending from Base event. Here is where we need to choose, bigger inheritance tree and smaller api or the other way round. But is totally possible if we want to.

android-core/src/main/java/com/mparticle/modernization/MParticleCallback.kt
markvdouw: Exactly, its a way of making callbacks more consistent across. Each callback will have either success (S) or error (E) type. In Android we will be introducing suspend functions (kind of promises) for async behavior, but since though this is the "new standard", callbacks have been there for ages and Java libraries cant use suspend functions and coroutines, we added support for a consistent way of using callbacks.

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
markvdouw: Yup exactly its just an option, we should follow a standard, either more precise and granular api or less granular but with more attributes)

android-core/src/main/java/com/mparticle/modernization/identity/MParticleIdentity.kt
markvdouw: Mostrly read/write access to stuff.. there are operations that should be internal only.

android-core/src/main/java/com/mparticle/modernization/identity/example/MParticleIdentityImpl.kt
markvdouw: Exactly, my ultimate idea is to do everything that makes sense and more performant to be done async, managed by the single coroutine thread pool in the Mediator. Because the implementation "XYZImpl" should be internal, it makes sense to do all the internal with a single way of handling async (threads and runnables, coroutines, etc.) We can then expose, wrap and provide further functionality in the external apis if needed.

android-core/src/main/java/com/mparticle/modernization/kit/MParticleKit.kt
markvdouw: Those are either listener interfaces or internal interfaces.. MParticleEventLogging is the api providing access to logEvent and those kind of things. Adding to it multiple interfaces can be implemented by this or any other class if needed.

android-core/src/main/java/com/mparticle/modernization/data/MParticleDataRepository.kt
markvdouw: The purpose is to have an abstraction managing data operation and multiple data sources if needed.
For example we migh want to retrieve something from a service (backendApi class), and while retrieving it, parallely save it in to a internal database and/or internal shared preferences to be used kind of a cache layer or so..
The advantange of the repository is just to make one call like "getUser" and hide the details of where the data came from, since all data sources should be injected, and business logic should handle it.

android-core/src/main/java/com/mparticle/modernization/uploading/MParticleUploadingStrategy.kt
```diff
@@ -0,0 +1,29 @@
+package com.mparticle.modernization.uploading
+
+import com.mparticle.modernization.config.UploadingConfiguration
+import org.jetbrains.annotations.NotNull
+import org.jetbrains.annotations.Nullable
+
+/**
+ * Data uploading strategies for different data types
+ */
+interface MParticleUploadingStrategy {
```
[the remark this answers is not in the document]
markvdouw: The idea of the Uploading strategy is to managing specific pre-processing of data before actually sending data to our servers. Currently we handle multiple types of data and our implementation have huge switch/if-else nested logic. The idea of this is to create multiple single responsibility data strategies for the different types. Also, if we pass a the manager class a collection with strategies and it only needs to pick the strategy, preprocess and send to the backend, we can abstract the UPLOADING component on its own.

android-core/src/main/java/com/mparticle/modernization/MParticleCallback.kt
```diff
@@ -0,0 +1,16 @@
...
+    var isSuccessFul: Boolean = false
+    var isCompleted: Boolean = false
+
+    fun onSuccess(result: S) {
+        isSuccessFul = true
```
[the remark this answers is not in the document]
markvdouw: Completed is kind of used for progressbars and those things.. if false implies that hasnt finished, while if true either success or error should be triggered

PR #359, opened 2023-04-05
ci: Change dependabot interval to monthly

jasonmgeorge:
## Instructions

 ## Summary
 - Change the dependabot execution interval from daily to monthly

 ## Testing Plan
 - Regression tests should pass

 - Closes https://go.mparticle.com/work/SQDSDKS-5302

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
     - package-ecosystem: gradle
       directory: "/"
       schedule:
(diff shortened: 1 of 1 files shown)


markvdouw (approved):

rmi22186 (approved):
I am not familiar with the `Run mobsfscan` actions which are failing, but I don't think this PR should be held up because of that.

PR #367, opened 2023-04-26
feat: modernization use case example (DRAFT-DO NOT MERGE)

markvdouw:
## Instructions

 ## Summary
 - Creating a use case code to outline the discussed architecture and show a real use case and interaction between layers.
This is obviously a WIP and is intendeed just to show high level interaction between the architecture components.

 ## Testing Plan


rmi22186 (commented):
Reviewed.  Looks good overall and looking forward to getting started on this project!

android-core/src/main/java/com/mparticle/modernization/BatchManager.kt
```diff
@@ -0,0 +1,13 @@
+package com.mparticle.modernization
+
+class BatchManager(private val api : MpApiClientImpl) {
```
einsteinx2, line 3: Same as other comment, this should take an `ApiClient` interface of which `MpApiClientImpl` extends right?

android-core/src/main/java/com/mparticle/modernization/core/MParticle.kt
```diff
@@ -0,0 +1,31 @@
+package com.mparticle.modernization.core
+
+import com.mparticle.MParticleOptions
+import com.mparticle.modernization.data.MParticleDataRepository
+import com.mparticle.modernization.eventlogging.MParticleEventLogging
+import com.mparticle.modernization.identity.MParticleIdentity
+import com.mparticle.modernization.kit.MParticleKitManager
+import com.mparticle.modernization.datahandler.MParticleDataHandler
+
+internal class MParticle private constructor(private val options: MParticleOptions) {
```
einsteinx2, line 10: What exactly does `private val` do in this constructor parameter? Is it syntax sugar for creating a private instance property called `options` and assigning it this value?

einsteinx2 (approved):
Left a couple comments and a question. Overall looks good though, I'll get a better feel for it once I port it to Swift.

android-core/src/main/java/com/mparticle/modernization/MpApiClientImpl.kt
```diff
@@ -0,0 +1,6 @@
+package com.mparticle.modernization
+
+class MpApiClientImpl {
```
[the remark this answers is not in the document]
markvdouw: Yup, just wanted to outline the idea of the data source itself. We might even have the ApiClient itself without interface if we don't intend to have multiple implementations of it (foe example multiple http clients as tools to resolve the implementation). In the future we could always create the abstraction and change this into an Impl.

android-core/src/main/java/com/mparticle/modernization/BatchManager.kt
```diff
@@ -0,0 +1,13 @@
+package com.mparticle.modernization
```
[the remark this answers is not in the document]
markvdouw: Sure

android-core/src/main/java/com/mparticle/modernization/core/MParticle.kt
markvdouw: This enables to use the dependency injection pattern, making the property accesible to be used in the class but restrict visibility from consumers. If options would have been public, MParticle.getInstance().options would have been valid.. Moreover by default properties have public setters/getters so, someone from the outside could eventually temper with the properties within MParticleOptions

jamesnrokt:
Closing as out of date

PR #368, opened 2023-04-27
fix:data plan blocking app crashes

mmustafa-tse:
## Summary
 - A customer reported that their app crashes when enabling blocking for their data plan, after investigating the error log shared, it seemed to be crashing at line 108 of the DataplanFilter.kt file. It seems function such as clear() and putAll() only work for mutableMaps and will crash of Kotlin's mapOf() is passed as a custom attribute. The fix here converts any incoming custom attribute maps

 ## Testing Plan
 - This was tested locally after creating a snapshot of the mParticle SDK that had the fix.

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5362


markvdouw:
LGTM

markvdouw (approved):

PR #378, opened 2023-05-08
feat: sideloading kits

markvdouw:
## Instructions

 ## Summary
Adding local kit as a wrapper for config and kit integration.
Creating empty config based on a random unused Int generated by the factory. 
Adding kit merging and initializing.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5112

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -15,7 +15,6 @@ import com.mparticle.internal.PushRegistrationHelper;
 import com.mparticle.networking.NetworkOptions;
 import com.mparticle.networking.NetworkOptionsManager;
 
-import org.jetbrains.annotations.NotNull;
 import org.json.JSONException;
 import org.json.JSONObject;
 
@@ -54,7 +53,7 @@ public class MParticleOptions {
     private MParticle.OperatingSystem mOperatingSystem = MParticle.OperatingSystem.ANDROID;
     private DataplanOptions mDataplanOptions;
     private Map<Class, List<Configuration>> mConfigurations = new HashMap();
-    private Map<Integer, Object> localKits = new HashMap<>();
+    private List<Object> sideloadedKits = new ArrayList<>();
 
     private MParticleOptions() {
     }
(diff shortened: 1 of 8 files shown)


android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -54,6 +54,7 @@ public class MParticleOptions {
     private MParticle.OperatingSystem mOperatingSystem = MParticle.OperatingSystem.ANDROID;
     private DataplanOptions mDataplanOptions;
     private Map<Class, List<Configuration>> mConfigurations = new HashMap();
+    private Map<Integer, Object> localKits = new HashMap<>();
```
einsteinx2, line 57: `localKits` should be called `sideloadedKits` to match the other platorms.

Also, why is this a map instead of an array like on the other platforms? Is there a specific need on Android for this?

android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -179,6 +181,16 @@ public MParticle.Environment getEnvironment() {
         return mEnvironment;
     }
 
+    /**
+     * Get list of local kits
```
einsteinx2, line 185: All uses of the term "local kits" in comments and variable names should be "sideloaded kits" to be in line with the other platforms

android-kit-base/src/main/java/com/mparticle/kits/KitIntegrationFactory.java
```diff
@@ -15,64 +16,82 @@
...
     final Map<Integer, Class> supportedKits = new HashMap<>();
+    final static Map<Integer, LocalKit> localKits = new HashMap<>();
 
-    public KitIntegrationFactory() {
+    public KitIntegrationFactory(MParticleOptions options) {
+        mergeIntegrations(options);
         loadIntegrations();
     }
 
+    public static int generateRandomKey() {
```
einsteinx2, line 26: Why generate random IDs? See the solution on iOS, we just defined a number (`1000000000`) that is the start of the sideloaded kits range, so that it's trivial to check if a kit is sideloaded or not just by checking it's ID without needing to reference a known list of IDs.

einsteinx2 (changes requested):

android-core/src/main/java/com/mparticle/MParticleOptions.java
markvdouw: Changed

PR #379, opened 2023-05-15
fix: SQDSDKS-5287 unable to drop batches nullability

markvdouw:
## Instructions

 ## Summary
 - Adding the ability to drop batches by returning null

 ## Testing Plan
 - Unit tests

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5287


einsteinx2 (changes requested):

einsteinx2 (approved):
LGTM thanks!

BrandonStalnaker (approved):
LGTM

PR #385, opened 2023-06-02
fix:screen attributes filters not working

mmustafa-tse:
## Instructions

 ## Summary
 - A customer reported that their screenView custom attributes were still forwarding even though it was filtered off on the UI, after further investigation and being able to repro I noticed in our code that we're passing the screenNames array to filter off instead of the screenAttibutes array.

 ## Testing Plan
 - [Y] Was this tested locally? If not, explain why.
 - Created a snapshot kit with the fix and tested locally


BrandonStalnaker (approved):
LGTM

mmustafa-tse:
Added unit testing for this case per @samdozor 's request

samdozor (approved):

PR #393, opened 2023-06-26
fix: Fixing initialization & de-initialization issue on kits.

markvdouw:
Fixing initialization & de-initialization issue on kits.

## Instructions

 ## Summary
 - Have and Android app with at least one integration kit connected and added as a dependency, re-run app after first install and you will see in the logs De-initializing kit: <kit name>. This happens quite often about 50% of the time.
Configuration was done pararelly for core & kits.
If any kit was present with a null or empty string (then converted to a JsonObject), the active kits were de-initialized and re-initialized because the preferences were overriden with empty string.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5408


einsteinx2 (approved):
Approved per Slack discussion. Defer to Mo for final e2e testing before merging.

BrandonStalnaker (approved):

PR #394, opened 2023-06-26
fix: Kit integration forwarding unplanned/blocked events when fired early on app run

markvdouw:
Kit integration forwarding unplanned/blocked events when fired early on app run

## Instructions

 ## Summary
 - Dataplans was being updated before the config has been loaded, resulting on a race condition.

 ## Testing Plan
@mmustafa-tse was able to verify the fix.

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5508

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
+++ b/android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
@@ -77,12 +77,13 @@ public class KitFrameworkWrapper implements KitManager {
                             .onKitsLoaded(() -> {
                                         mKitManager = kitManager;
                                         setKitsLoaded(true);
+                                        updateDataplan(mCoreCallbacks.getDataplanOptions());
                                     }
                             );
                 } else {
                     mKitManager = kitManager;
+                    updateDataplan(mCoreCallbacks.getDataplanOptions());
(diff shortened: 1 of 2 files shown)


android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
```diff
@@ -77,12 +77,13 @@ public void loadKitLibrary() {
                             .onKitsLoaded(() -> {
                                         mKitManager = kitManager;
                                         setKitsLoaded(true);
+                                        updateDataplan(mCoreCallbacks.getDataplanOptions());
```
samdozor, line 80: this reads like a race condition - we need to ensure that the data plan is set before kits are loaded. specifically - we need to make sure that no user attribute, events, etc can be sent to kits prior to the data plan.

in my view ideally we'd load the data plan in the same function `updateKits` above that loads kit configuration. the idea of KitConfiguration is that it should contain absolutely everything needed, and we only start kits after it's been completely parsed.

samdozor (changes requested):

samdozor (approved):

BrandonStalnaker (approved):
LGTM

PR #398, opened 2023-07-03
fix: Kit Initialization & De-Initialization.

markvdouw:
Fix for Kit Initialization & De-Initializationg.

## Instructions

 ## Summary
The underlying issues is that when the KitFrameworkManager calls fetchConfig, the kitConfig instance is null (the instance created not yet assigned). Therefore, the ApiClient can get the "supported kits" from the [KitManager.KitIntegrationFactory] and the mp-kit header is not added to the call, resulting in a response with "eks" empty. 

The fix includes assigning the KitManager instance before calling configure kits (which will fetch the config). 
The kitManager config will anyway be reassigned in the configuration callback.
Removing empty validation in ConfigManager since this is a valid case.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5408

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/ConfigManager.java
@@ -319,12 +319,10 @@ public class ConfigManager {
                     .putString(Constants.PrefKeys.ETAG, etag)
                     .putString(Constants.PrefKeys.IF_MODIFIED, lastModified)
                     .apply();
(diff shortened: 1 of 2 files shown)


samdozor (approved):

schallock:
@einsteinx2 I have a tech writing ticket ([TW-1204](https://mparticle-eng.atlassian.net/browse/TW-1204)) that was automatically created for this bug - from from what I can tell, there's no actual doc impact. Can I mark my ticket as "won't do", or am I missing something?

PR #401, opened 2023-07-06
feat: sideloading kits

markvdouw:
## Instructions

 ## Summary
 - Adding kit filtering to MPSidelodingKit abstraction which extends from KitIntegration itself
 - Adding SideloadedKit interface
 - Adding sideloaded kits to MParticleOptions and its builder
 - Merging remote config with manually passed sideloaded kits in MParticleOptions (considering cases of api responses 200-300 and 304) expecting remote configurations with known integrations, and, sideloaded kits to work exctly the same (lifecycle, logging, initialization & de-initialization, etc)
 - Adding some merging logic in ConfigManager which saves and retrieves processed config into/from shared prefs.
 - Adding unit tests
 - Adding param to DeviceAttributes to report number of sideloaded kits
 - Integration tested with Higgs sample app.

 ## Testing Plan
 - [x] Was this tested locally? If not, explain why. (Unit testing and functional testing on HiggsApp)

 - Closes:
https://mparticle-eng.atlassian.net/browse/SQDSDKS-5115
https://mparticle-eng.atlassian.net/browse/SQDSDKS-5390
https://mparticle-eng.atlassian.net/browse/SQDSDKS-5484


android-core/src/main/java/com/mparticle/internal/DeviceAttributes.java
```diff
@@ -42,9 +46,20 @@ public static String getDeviceImei() {
...
+                count++;
+            }
+        }
+        return count;
+    }
+
     /**
      * Generates a collection of application attributes that will not change during an app's process.
-     *
+     * <p>
```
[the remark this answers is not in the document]
markvdouw: mm let me review.

android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -421,6 +434,18 @@ public Builder installType(@NonNull MParticle.InstallType installType) {
             return this;
         }
 
+        /**
+         * Add locsideloadedKitsal kits
```
einsteinx2, line 438:
```suggestion
         * Add sideloaded kits
```

Looks like a copy/paste error here?

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -277,7 +283,7 @@ void checkConfigStaleness() {
 
     @Nullable
     String getConfig() {
-        return sPreferences.getString(CONFIG_JSON, null);
+        return sPreferences.getString(CONFIG_JSON, "");
```
einsteinx2, line 286: I assume this change means that this will return a blank string instead of null if the config doesn't exist, but the method is marked nullable... Since this is an internal class, can we mark this non null since it will now always return a string?

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
```diff
@@ -0,0 +1,90 @@
+package com.mparticle.kits
```
einsteinx2, line 1: If this file isn't needed, we should remove it from the PR or just include an empty class definition if the type is used.

This class has things with comments like "web only" or "iOS only" that should not be merged into development/main.

einsteinx2 (changes requested):

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -114,16 +114,200 @@ public SparseBooleanArray getScreenAttributeFilters() {
...
+        return object;
+    }
+
+    private static JSONObject convertSparseMapToJsonObject(Map<Integer, String> map) throws JSONException {
+        if (map == null || map.size() == 0) {
+            return null;
+        }
+        JSONObject obj = new JSONObject();
+        for (Map.Entry<Integer, String> entry : map.entrySet()) {
+//            try {
```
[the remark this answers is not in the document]
markvdouw: Originally the idea was: because the MParticleSideloadedFilters were not implemented, this will never throw an exception yet, but by the time of implementing it I didnt want to catch Exception without knowing that something happened, so that's why I added the try/catch and commented as if it was a "TODO" comment, but I can change it for a TODO

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -114,16 +114,200 @@ public SparseBooleanArray getScreenAttributeFilters() {
...
+
+    public static JSONObject convertSparseArrayToJsonObject(SparseBooleanArray array) throws JSONException{
+        if (array == null || array.size() == 0) {
+            return null;
+        }
+        JSONObject object = new JSONObject();
+        for (int i = 0; i < array.size(); i++) {
+            int key = array.keyAt(i);
+            //TODO Should parse to String?
+//            try {
```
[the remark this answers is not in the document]
markvdouw: Idem

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -114,16 +114,200 @@ public SparseBooleanArray getScreenAttributeFilters() {
...
+    }
+
+    public static JSONObject convertSparseArrayToJsonObject(SparseBooleanArray array) throws JSONException{
+        if (array == null || array.size() == 0) {
+            return null;
+        }
+        JSONObject object = new JSONObject();
+        for (int i = 0; i < array.size(); i++) {
+            int key = array.keyAt(i);
+            //TODO Should parse to String?
```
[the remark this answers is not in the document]
markvdouw: Idem, this feature is outlined but not implemented, it was just to keep track of this for the future.

android-core/src/main/java/com/mparticle/MParticleOptions.java
markvdouw: Yup, Ill fix it.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
markvdouw: Ill review it, I think yes, but because its public I want to double-check it.

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
markvdouw: The class is needed as the new MParticleSideloaded abstraction has a fuction to addFilters(filters : MPSideloadedFilters) required on tje Jira ticket.

We can delete the comments, but thats all.

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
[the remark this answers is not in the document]
einsteinx2: Yeah as a rule we don't check in commented out code. A todo comment is ok as long as you remember to get back to it when finishing the work.

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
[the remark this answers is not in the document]
einsteinx2: Ok that's fine.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
einsteinx2: if it's public, why is it inside a package called "internal"?

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
einsteinx2: Actually sorry I should have been clearer in the Jira ticket. It means to add the MPSideloaded kit class only as far as it needs to be implemented to not have to change the public API later. None of the addFilter methods actually work yet, so they shouldn't be there. See the example in the iOS code base, it only has the kit reference and nothing else: https://github.com/mParticle/mparticle-apple-sdk/blob/main/mParticle-Apple-SDK/Kits/MPSideloadedKit.swift

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
markvdouw: MParticleOptions has always been in the internal package which in my opinion actually menas "core" (like where the core logic is). Most of the things, specifically in the Builder class needs to be public to actually be able to build the object.

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
markvdouw: OK Ill do that.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
einsteinx2: Gotcha, then yeah we can't change the API then, agreed.

einsteinx2 (approved):
Looks good thanks!

PR #408, opened 2023-07-25
feat: Adding filtering hashing functions

markvdouw:
## Instructions

 ## Summary
 - Moving filtering hashing functions to utility dedicated class and replacing old ones in the code.
 - Changing tests
 - All tests must pass since the logic wasn't change, just ones in a different class function

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5346


android-core/src/main/java/com/mparticle/internal/HashingUtility.kt
```diff
@@ -0,0 +1,66 @@
...
+ */
+object HashingUtility {
+
+    fun hashEvent(event: MPEvent): Int =
+        MPUtility.mpHash("${event.eventType.ordinal}${event.eventName}")
+
+    fun hashMpMessageTypeName(mpMessage: BaseMPMessage): Int =
+        MPUtility.mpHash("${mpMessage.messageType}${mpMessage.name}")
+
+    fun hashFilterAvKey(key: String): Int = MPUtility.mpHash(key)
```
BrandonStalnaker, line 22: Why split these up into two (hashFilterAvKey & hashFilterAvValue) separate methods? Seems like an unnecessary complexity.

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -610,10 +610,10 @@ boolean shouldIncludeFromAttributeValueFiltering(Map<String, String> attributes)
                 while (attIterator.hasNext()) {
                     Map.Entry<String, String> entry = attIterator.next();
                     String key = entry.getKey();
-                    int keyHash = KitUtils.hashForFiltering(key);
+                    int keyHash = HashingUtility.INSTANCE.hashFilterAvKey(key);
```
BrandonStalnaker, line 613: I'm not sure if I seen this pattern before in android. Would it be possible to remove INSTANCE or at least change it to something that matches the capitalization style of the rest of the code.

BrandonStalnaker (commented):
Seems good in general. Just have a few suggestions/questions

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
markvdouw: HashingUtility is an object class in Kt (similar to a class level companion object that correspond to static).
Because functions are not called using an instance the INSTANCE wouldn't be needed if we were calling if from a Kt class also, but because KitConfiguration is a java class that is mandatory and also automatic.

android-core/src/main/java/com/mparticle/internal/HashingUtility.kt
markvdouw: Ben asked me to create functions in the utility class based on semantic so because conceptually there could be different ways to hash those (since conceptually those are different things) we wanted to have the separately event though the implementation is the same.

BrandonStalnaker (approved):
LGTM but I'd prefer if we also got @einsteinx2 approval too since he knows the implementation you're following best

android-core/src/main/java/com/mparticle/internal/HashingUtility.kt
einsteinx2: Yes that's correct, it's some slight code duplication but makes the purpose clearer at the call site. Plus if we ever change the hash method in the future (unlikely but possible) it's more flexible, but the main reason is the former.

android-kit-base/src/test/kotlin/com/mparticle/kits/KitConfigurationTest.kt
```diff
@@ -450,7 +450,7 @@ class KitConfigurationTest {
         consentForwardingRule.put("v", ruleArray)
         val rule1 = JSONObject()
         ruleArray.put(rule1)
-        rule1.put("h", KitUtils.hashForFiltering("1" + "foo purpose"))
+        rule1.put("h", HashingUtility.hashGDPRContentPurposeKey("1" + "foo purpose"))
```
einsteinx2, line 453:
```suggestion
        rule1.put("h", HashingUtility.hashGDPRContentPurposeKey("foo purpose"))
```
Shouldn't the `"1" +` be removed here since the new method adds it?

android-kit-base/src/test/kotlin/com/mparticle/kits/KitConfigurationTest.kt
```diff
@@ -571,7 +585,7 @@ class KitConfigurationTest {
         consentForwardingRule.put("v", ruleArray)
         val rule1 = JSONObject()
         ruleArray.put(rule1)
-        rule1.put("h", KitUtils.hashForFiltering("1" + "foo purpose"))
+        rule1.put("h", HashingUtility.hashGDPRContentPurposeKey("foo purpose"))
```
einsteinx2, line 588: For example here the `"1" +` was removed...

In that case, why did the previous error not cause a test failure? Are we not actually testing that data? If not, why include it in the test?

einsteinx2 (changes requested):

einsteinx2 (changes requested):

PR #413, opened 2023-07-31
feat: filtering dictionary parsing, hashing functions and tests for sideloading kits filtering feature

markvdouw:
## Instructions

 ## Summary
 - Filtering dictionary parsing, hashing functions and tests for sideloading kits filtering feature
Merging this into filtering feature branch to merge it to develop all together.
Note that the JSON created for testing local filter creation can be used in other platforms.
Merging feat/SQDSDKS-5346 branch into this (already reviewed https://github.com/mParticle/mparticle-android-sdk/pull/408)

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5346
 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5347
 -  Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5348


android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
```diff
@@ -1,9 +1,297 @@
...
+     *
+     * @param userIdentityType
+     * @return MPSideloadedFilters instance
+     */
+    // Internally use uid as the key
+    fun addUserIdentityFilter(userIdentityType: MParticle.IdentityType): MPSideloadedFilters {
+        applyToFiltersNode(
+            KEY_USER_IDENTITY_FILTER,
+            Pair(
+                userIdentityType.value.toString(),
```
einsteinx2, line 144: So even though user identity type isn't actually hashed, for consistency sake, the plan is to include it in the hashing utility class and just return back the value. See the web implementation here: https://github.com/mParticle/mparticle-web-sdk/blob/ea26d51e8c09129d9550f937192208a5368f75e2/src/kitFilterHelper.ts#L25

The idea is to keep it consistent, allow for potential hashing in the future, and self-document that in this case the "hash" is just the value itself.

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
```diff
@@ -1,9 +1,297 @@
...
+     */
+    // Internally use svea as the key
+    fun addScreenAttributeFilter(
+        screenName: String,
+        customAttributeKey: String
+    ): MPSideloadedFilters {
+        applyToFiltersNode(
+            KEY_SCREEN_ATTRIBUTES_FILTER,
+            Pair(
+                HashingUtility.hashFilterScreenName("$screenName$customAttributeKey").toString(),
```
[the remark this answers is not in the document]
markvdouw: Creating a new hashing fun

android-kit-base/src/main/java/com/mparticle/kits/MPSideloadedFilters.kt
markvdouw: Creating a hashing fun

PR #431, opened 2023-09-29
feat: Adding pod specific redirection based on prefix

markvdouw:
## Instructions

 ## Summary
 - Changes for pod specific redirection

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5770

### DIFF
--- a/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
+++ b/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
@@ -20,7 +20,7 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
         startMParticle(MParticleOptions.builder(mContext).credentials(apiKey, "secret"))
         val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
         for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
-            defaultUrls[endpoint] = baseClientImpl.getUrl(endpoint, endpoint.name)
+            defaultUrls[endpoint] = baseClientImpl.getUrl(endpoint, endpoint.name, false)
         }
         MParticle.setInstance(null)
     }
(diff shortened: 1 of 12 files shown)


android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -54,11 +54,11 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
             .build()
         MParticle.start(options)
         val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
-        for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
-            val generatedUrl = baseClientImpl.getUrl(endpoint, endpoint.name, true)
-            Assert.assertEquals(defaultUrls[endpoint].toString(), generatedUrl.toString())
-            TestCase.assertTrue(generatedUrl === generatedUrl.defaultUrl)
-        }
+//        for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
```
einsteinx2, line 57: We shouldn't merge commented out stuff, all tests should pass with the new implementation

android-core/src/main/java/com/mparticle/networking/NetworkOptionsManager.java
```diff
@@ -11,23 +11,23 @@
 
 public class NetworkOptionsManager {
     public static String MP_CONFIG_URL = "config2.mparticle.com";
-    public static String MP_IDENTITY_URL = "identity.mparticle.com";
-    public static String MP_URL = "nativesdks.mparticle.com";
+    public static String MP_IDENTITY_URL = "identity";
```
einsteinx2, line 14: Nitpick but for clarity we should rename these to something like `MP_IDENTITY_SUBDOMAIN` or `MP_IDENTITY_URL_PREFIX` to clarify that it's not actually a URL. I did something similar on iOS

einsteinx2 (changes requested):
Couple of small things, otherwise looks good, I see we went in different directions on the implementation but the overall network implementations on iOS and Android are different so that's fine.

PR #438, opened 2023-10-30
feat: SQDSDKS-5914- Server redirection based on apikey prefix and boolean config flag

markvdouw:
## Instructions

 ## Summary
 - Server redirection based on config boolean flag and api-key prefix.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5914


einsteinx2 (approved):
Now that we confirmed config parsing this is good to go.

PR #447, opened 2023-11-06
feat: SQDSDKS-5914 - Pod redirection

markvdouw:
## Instructions

 ## Summary
 - Pod redirection feature branch merge

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5914

### DIFF
--- a/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
+++ b/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
@@ -10,6 +10,7 @@ import org.junit.Assert
 import org.junit.Before
 import org.junit.Test
 import java.net.MalformedURLException
+import kotlin.test.assertEquals
 
 class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
     var defaultUrls = HashMap<MParticleBaseClientImpl.Endpoint, MPUrl>()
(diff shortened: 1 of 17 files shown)


android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -20,11 +21,25 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
         startMParticle(MParticleOptions.builder(mContext).credentials(apiKey, "secret"))
         val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
         for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
-            defaultUrls[endpoint] = baseClientImpl.getUrl(endpoint, endpoint.name)
+            defaultUrls[endpoint] = baseClientImpl.getUrl(endpoint, endpoint.name, false)
         }
         MParticle.setInstance(null)
     }
 
+    @Test
```
einsteinx2, line 29: We should add more comprehensive testing. Looks like this only tests the final URL creation, not ensuring it pulls the correct prefix from different keys, and doesn't test all pods or keys without prefixes.

See the iOS tests here for reference: https://github.com/mParticle/mparticle-apple-sdk/blob/14168f9167f15f8b850b93ea1004273cf927bd0e/UnitTests/MPNetworkCommunicationTests.m#L841

Web also tests similarly to iOS.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1276,6 +1280,20 @@ public ConsentState getConsentState(long mpid) {
         return ConsentState.withConsentState(serializedConsent).build();
     }
 
+    public boolean podRedirectionEnabled() {
+        return directUrlRouting;
+    }
+
+    public String getPodPrefix() {
+        String prefix = "us1";
+        try {
```
einsteinx2, line 1289: Isn't it generally bad practice to use exceptions for control flow? I think it would be better to just call split, then check the length with an if statement rather than relying on the exception to fall back to the us1 case...

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -110,6 +111,8 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
             url = domainMappingUrl;
         }
         Uri uri;
+        if(endpoint!=Endpoint.CONFIG){
```
einsteinx2, line 114:
```suggestion
        if (endpoint != Endpoint.CONFIG ) {
```
Nitpick, we should fix the spacing here.

einsteinx2 (changes requested):

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -51,6 +51,7 @@ public class ConfigManager {
     static final String KEY_OPT_OUT = "oo";
     public static final String KEY_UNHANDLED_EXCEPTIONS = "cue";
     public static final String KEY_PUSH_MESSAGES = "pmk";
+    public static final String DIRECT_URL_ROUTING = "dur";
```
jasonmgeorge, line 54: Any reason to deviate from the "KEY" nomenclature used for these?

```suggestion
    public static final String KEY_DIRECT_URL_ROUTING = "dur";
```

rmi22186:
closing in favor of https://github.com/mParticle/mparticle-android-sdk/pull/499

PR #450, opened 2023-11-13
feat: Background Event batching based delay

markvdouw:
Background Event batching based delay

## Instructions

 ## Summary
- Adding JobScheduler utilities with current and legacy code to trigger batch upload
- Implementing as the legacy impl, a delayed message with type UPLOAD based on dynamic delay 
- Adding feature flag ebb to "enableBackgroundBatching", read from config and default=false Adding Service for uploading
- Adding class in proguard
- Business logic change at the upload handler level to dismiss automatic upload interval and upload at session end when batching is enabled.

 ## Testing Plan
 Please consider that the JobScheduler specifies that the job will be triggered at MOST one time within the interval but there is no guarantee when the job will be trigger, meaning the delay can be very inexact and this could be a challange while testing.

Images are attached:
Screen Shot 2023-12-06 at 13 33 52

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5924 and https://mparticle-eng.atlassian.net/browse/SQDSDKS-5925

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
@@ -17,6 +17,7 @@ import androidx.annotation.Nullable;
 import com.mparticle.JobSchedulerUtilsKt;
 import com.mparticle.MPEvent;
 import com.mparticle.MParticle;
+import com.mparticle.SchedulingBatchingType;
 import com.mparticle.identity.IdentityApi;
 import com.mparticle.identity.IdentityApiRequest;
 import com.mparticle.identity.MParticleUser;
@@ -377,13 +378,11 @@ public class AppStateManager {
     }
 
     private void scheduleBackgroundJob() {
-        if (mConfigManager.isBackgroundEventBatchingEnabled()) {
-            JobSchedulerUtilsKt.scheduleBatchUploading(this.mContext, delay -> {
-                mMessageManager.mUploadHandler.sendMessageDelayed(mMessageManager.mUploadHandler.obtainMessage(UploadHandler.UPLOAD_TRIGGER_MESSAGES, 1, 0, mConfigManager.getMpid()), delay);
-                Logger.debug("Legacy action with delay: " + delay);
-                return Unit.INSTANCE;
-            });
-        }
+        JobSchedulerUtilsKt.scheduleBatchUploading(this.mContext, mConfigManager.getUploadInterval(), SchedulingBatchingType.ONE_SHOT, true, delay -> {
+            mMessageManager.mUploadHandler.sendMessageDelayed(mMessageManager.mUploadHandler.obtainMessage(UploadHandler.UPLOAD_TRIGGER_MESSAGES, 1, 0, mConfigManager.getMpid()), delay);
+            Logger.debug("Legacy action with delay: " + delay);
+            return Unit.INSTANCE;
+        });
     }
 
     @TargetApi(14)
--- a/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
@@ -136,6 +136,7 @@ public class AppStateManager {
 
     public void onActivityResumed(Activity activity) {
         try {
+            scheduleBackgroundJob();
             mCurrentActivityName = AppStateManager.getActivityName(activity);
 
             int interruptions = mInterruptionCount.get();
@@ -372,13 +373,12 @@ public class AppStateManager {
             instance.Internal().getKitManager().onApplicationBackground();
             mCurrentActivityName = null;
             Logger.debug("App backgrounded.");
-            scheduleBackgroundJob();
             mInterruptionCount.incrementAndGet();
         }
     }
 
     private void scheduleBackgroundJob() {
-        JobSchedulerUtilsKt.scheduleBatchUploading(this.mContext, mConfigManager.getUploadInterval(), SchedulingBatchingType.ONE_SHOT, true, delay -> {
+        JobSchedulerUtilsKt.scheduleBatchUploading(this.mContext, mConfigManager.getUploadInterval(), SchedulingBatchingType.ONE_SHOT, delay -> {
             mMessageManager.mUploadHandler.sendMessageDelayed(mMessageManager.mUploadHandler.obtainMessage(UploadHandler.UPLOAD_TRIGGER_MESSAGES, 1, 0, mConfigManager.getMpid()), delay);
             Logger.debug("Legacy action with delay: " + delay);
             return Unit.INSTANCE;
--- a/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/AppStateManager.java
@@ -14,6 +14,7 @@ import android.os.SystemClock;
 
 import androidx.annotation.Nullable;
 
+import com.mparticle.JobSchedulerUtilsKt;
 import com.mparticle.MPEvent;
 import com.mparticle.MParticle;
 import com.mparticle.identity.IdentityApi;
@@ -26,6 +27,8 @@ import org.json.JSONObject;
 import java.lang.ref.WeakReference;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+
+import kotlin.Unit;
 
 
 /**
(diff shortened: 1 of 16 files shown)


android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -922,9 +924,11 @@ public boolean shouldTrigger(BaseMPMessage message) {
             isBackgroundAst = (message.getMessageType().equals(Constants.MessageType.APP_STATE_TRANSITION) && message.get(Constants.MessageKey.STATE_TRANSITION_TYPE).equals(Constants.StateTransitionType.STATE_TRANS_BG));
         } catch (JSONException ex) {
         }
+        if (isBackgroundAst) {
+            return false;
+        }
         boolean shouldTrigger = message.getMessageType().equals(Constants.MessageType.PUSH_RECEIVED)
-                || message.getMessageType().equals(Constants.MessageType.COMMERCE_EVENT)
-                || isBackgroundAst;
+                || message.getMessageType().equals(Constants.MessageType.COMMERCE_EVENT);//|| isBackgroundAst;
```
einsteinx2, line 931: `//|| isBackgroundAst;`

We should get rid of this extra commented out code

android-core/src/main/java/com/mparticle/internal/UploadHandler.java
```diff
@@ -145,8 +151,19 @@ public void handleMessageImpl(Message msg) {
                             }
                         }
                     }
+                    if (!anythingToUpload && (!mAppStateManager.getSession().isActive() || mAppStateManager.isBackgrounded())) {
+                        //If there is anything to upload and is in background or inactive, cancel schedule job until a new message is stored
```
einsteinx2, line 155: Should this say "//If there is **not** anything to upload"? or is the check for `!anythingToUpload` in the if statement incorrect?

android-core/src/main/java/com/mparticle/internal/AppStateManager.java
```diff
@@ -368,10 +372,19 @@ private void logBackgrounded() {
             instance.Internal().getKitManager().onApplicationBackground();
             mCurrentActivityName = null;
             Logger.debug("App backgrounded.");
+            scheduleBackgroundJob();
             mInterruptionCount.incrementAndGet();
         }
     }
 
+    private void scheduleBackgroundJob() {
```
einsteinx2, line 380: What is the purpose of this separate task vs just using the repeating one? Weren't we going to merge them into one or am I misremembering the details?

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -407,6 +408,7 @@ private synchronized void updateCoreConfig(JSONObject responseJSON, boolean newC
         }
 
         mRampValue = responseJSON.optInt(KEY_RAMP, -1);
+//        isBackgroundEventBatchingEnabled = responseJSON.optBoolean(ENABLE_BACKGROUND_BATCHING, false);
```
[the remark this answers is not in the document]
markvdouw: Done

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
markvdouw: Done

android-core/src/main/java/com/mparticle/internal/UploadHandler.java
markvdouw: Good catch

android-core/src/main/java/com/mparticle/internal/AppStateManager.java
markvdouw: This function is scheduling the job.
In terms why ONE_SHOT vs PERIODIC, the idea is to start a ONE_SHOT when going on background to prevent recurrently uploading stuff with no need. When the ONE_SHOT triggers we will set a recurrent one based on the uploadInterval, being active and having stuff to send. Its a combination of both.

android-core/src/main/kotlin/com.mparticle/JobSchedulerUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.annotation.RequiresApi
+import com.mparticle.internal.Logger
+import com.mparticle.uploadbatching.BatchUploadingJob
+
+const val UPLOAD_BATCH_JOB = 123
+
+fun Long.secondsToMillis(): Long = this * 1000
+fun Long.minutesToMillis(): Long = (this * 60).secondsToMillis()
+
+enum class SchedulingBatchingType { PERIODIC, ONE_SHOOT; }
```
einsteinx2, line 17: Just noticed that `ONE_SHOOT` should be `ONE_SHOT` :)

einsteinx2 (changes requested):

android-core/src/main/java/com/mparticle/internal/AppStateManager.java
einsteinx2: Sounds good, thanks for the explanation.

android-core/src/main/kotlin/com.mparticle/JobSchedulerUtils.kt
einsteinx2: I'm pretty sure after this change I can approve.

android-core/src/main/java/com/mparticle/internal/UploadHandler.java
```diff
@@ -145,8 +151,19 @@ public void handleMessageImpl(Message msg) {
...
+                        JobSchedulerUtilsKt.cancelScheduledUploadBatchJob(mContext);
+                    }
                     if (mAppStateManager.getSession().isActive() && uploadInterval > 0 && msg.arg1 == 0) {
-                        this.sendEmptyDelayed(UPLOAD_MESSAGES, uploadInterval);
+                        JobSchedulerUtilsKt.scheduleBatchUploading(mContext,
+                                uploadInterval,
+                                SchedulingBatchingType.PERIODIC,
+                                true,
+                                delay -> {
+                                    sendEmptyDelayed(UPLOAD_MESSAGES, uploadInterval);
```
samdozor, line 164: there's something funky here. I tested by setting the upload interval to 30 seconds and set the SDK to production mode. When using the legacy upload mode -  i'm not seeing uploads every 30s - it seems to be much quicker.

android-core/src/main/kotlin/com.mparticle/JobSchedulerUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+            } else if (type == SchedulingBatchingType.ONE_SHOT) {
+                builder.setMinimumLatency(delayInSeconds.secondsToMillis())
+            }
+            builder.apply {
+                setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY)
+            }
+            context.scheduleJob(builder.build())
+        } catch (e: Exception) {
+            Logger.error("Error while trying to use JobSceduler for batch upload")
+            legacyAction.invoke(delayInSeconds.secondsToMillis())
```
samdozor, line 57: delayInSeconds is already in milliseconds when it's passed in....

android-core/src/main/kotlin/com.mparticle/JobSchedulerUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+                builder.setPeriodic(delayInSeconds.secondsToMillis())
+            } else if (type == SchedulingBatchingType.ONE_SHOT) {
+                builder.setMinimumLatency(delayInSeconds.secondsToMillis())
+            }
+            builder.apply {
+                setRequiredNetworkType(JobInfo.NETWORK_TYPE_ANY)
+            }
+            context.scheduleJob(builder.build())
+        } catch (e: Exception) {
+            Logger.error("Error while trying to use JobSceduler for batch upload")
```
samdozor, line 56: When using this I get spammed with error messages saying "Error while trying to use JobSceduler for batch upload" (including the typo..).

Can we check programmatically if the service is there rather than relying on an exception?

And - rather than a vague error, we should log more or a warning that includes the code that they need to add to their manifest, and let them know that as a result of this, we may not be able to upload in the background.

samdozor (changes requested):
I haven't tested yet with the job scheduler working - here's what I found if a customer just upgrades to the SDK. 


As part of this PR - we need to update the README of the SDK to have the docs for how to setup the scheduler.

PR #452, opened 2023-11-17
feat: SQDSDKS-5855 - Identity caching

markvdouw:
Checking if we should make an identiy call or return saved MParticleUser based on last request (using timeout) and several conditions.

## Instructions

 ## Summary
 - Adding logic to check if an identity request should be done or no based on mpId, userIdentities and params such as deviceId and googleId (implementation needed) as well as timeout and lastTime request was made.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5975


thomson-t:
Closing as this branch is outdated.

PR #461, opened 2023-12-11
feat: SQDSDKS-5924 - Batching events with delay after app goes backgrounded

markvdouw:
## Instructions

 ## Summary
Whenever the upload action is receive by our handlers and the batches are upload, we trigger a next upload message if the uploadInterval set by the user is greater than 0 and either 
-The app isNotBackgrounded and the session is active  OR
-The app isBackgrounded and either any of the batch upload failed and we would like to retry it in the future or there are messages ready to be sent in the future

When going to background an alarm is set based on the uploadInterval to upload batches even if the app is closed.
If the app is switched between background and foreground before scheduling a new alarm the previous one will be canceled.

 ## Testing Plan

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5924

Screen Shot 2023-12-11 at 07 37 11


thomson-t:
Outdated. Closing the PR

PR #463, opened 2023-12-13
fix: SQDSDK-5597 - Get ad id in background thread

markvdouw:
## Instructions

 ## Summary
 - Move the getAdIdInfo to be triggered in a background thread

 ## Testing Plan
 - Added a log printing the current thread in which the fun was being called (which was then removed prior to commit) 
Screen Shot 2023-12-13 at 09 11 52

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-5867

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/MPUtility.java
+++ b/android-core/src/main/java/com/mparticle/internal/MPUtility.java
@@ -720,11 +720,6 @@ public class MPUtility {
             if (caseInsensitive) {
                 key = findCaseInsensitiveKey(attributes, key);
             }
-
-            if (Constants.LIMIT_ATTR_COUNT == attributes.length() && !attributes.has(key)) {
-                Logger.error("Attribute count exceeds limit. Discarding attribute: " + key);
-                return false;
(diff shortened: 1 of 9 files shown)


android-core/src/main/java/com/mparticle/internal/MPUtility.java
```diff
@@ -20,6 +20,7 @@
 import android.os.StatFs;
 import android.provider.Settings;
 import android.telephony.TelephonyManager;
+import android.util.Log;
```
samdozor, line 23:
```suggestion
```

samdozor (changes requested):
would you mind also annotating the getAdIdInfo method with  @workerthread (or whatever the syntax is)?

markvdouw:
> would you mind also annotating the getAdIdInfo method with @workerthread (or whatever the syntax is)?

This is already done

markvdouw (commented):
Already has the @WorkerThread annotation in line 128

samdozor (approved):

PR #491, opened 2024-05-29
fix: Refactor code for accessing Google Ad info

Mansi-mParticle:
## Instructions

 ## Summary
 - Currently, the current thread will not block each time to read the status of Google Ad information. It only blocks for the first time; subsequently, it updates on a separate thread.

 ## Testing Plan
 - Tested locally and with new unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6465


einsteinx2 (approved):
Discussed approach before starting the work, implementation looks good to me.

samdozor (approved):
looks ok. doesn't need to hold up this PR but there must be some cleaner way to structure our access to the google ad id in the future. We only need to update the google ad id once per foreground (it literally cannot change more than that), and I see that we're calling this api all over the place. in the long run i suspect better to design a class that only updates on foreground, and have all other consumers get it from that class.

PR #492, opened 2024-05-30
fix: Add callback functionality to upload method

Mansi-mParticle:
## Instructions

 ## Summary
 - Added callback functionality to the upload method. The callback parameter is only required when using it with the 'doUpload()' method in the 'MessageManager' class.

 ## Testing Plan
 - Tested locally and with new unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6465

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/MessageManager.java
+++ b/android-core/src/main/java/com/mparticle/internal/MessageManager.java
@@ -62,6 +62,7 @@ public class MessageManager implements MessageManagerCallbacks, ReportingManager
     private ConfigManager mConfigManager = null;
     private MParticleDBManager mMParticleDBManager;
     private MParticle.OperatingSystem mOperatingSystem;
+    private MParticle.UploadCallback uploadCallback;
 
 
     /**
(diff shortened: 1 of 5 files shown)


android-core/src/main/java/com/mparticle/internal/MessageManager.java
```diff
@@ -525,7 +526,10 @@ public void startUploadLoop() {
         mUploadHandler.sendMessageDelayed(mUploadHandler.obtainMessage(UploadHandler.UPLOAD_MESSAGES, mConfigManager.getMpid()), Constants.INITIAL_UPLOAD_DELAY);
     }
 
-    public void doUpload() {
+    public void doUpload(MParticle.UploadCallback... callbacks) {
```
einsteinx2, line 529: So is the idea of using varargs parameter so that you can call this without a callback so that we don't break the existing public API? I think this is a bit confusing because it implies you can have multiple callbacks but then only uses the first one...

I think it would be better to leave the original `public void doUpload()` method as-is, then add a second method like so:

```
public void doUpload(MParticle.UploadCallback callback) {
    uploadCallback = callback;
    doUpload();
}
```

That should make it clearer without breaking the existing public API.

einsteinx2 (changes requested):

einsteinx2 (approved):
lgtm

einsteinx2:
No longer relevant with new workspace switching implementation

PR #495, opened 2024-06-25
feat: Add option to disable certificate pinning

Mansi-mParticle:
## Instructions

 ## Summary
 - Added an option to disable certificate pinning, even in production.

 ## Testing Plan
 - Tested with sample application
 - Bad certificate and pinning option is false ( no response from the api)
 
Screenshot 2024-06-25 at 5 30 58 PM

 - Bad certificate and pinning option is true 
 
Screenshot 2024-06-25 at 5 33 37 PM

Screenshot 2024-06-25 at 5 34 55 PM

 - Closes https://go.mparticle.com/work/SQDSDKS-6541


einsteinx2 (approved):
Looks good, thanks for the detailed verification support!

PR #496, opened 2024-07-08
fix: Add enum value for event types

Mansi-mParticle:
## Instructions

 ## Summary
 - Add enum values for event types

 ## Testing Plan
 - Tested with sample application and executed unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6544


rmi22186 (approved):
LGTM

PR #498, opened 2024-07-10
fix: Add enum types for commerce event

Mansi-mParticle:
## Instructions

 ## Summary
 - Added enums for e-commerce events.
 - 'expandProductAction' to create an event instead of a transaction. Separate it into different e-commerce event types.

 ## Testing Plan
 - Tested it with a sample application and ran unit tests

 - Closes https://go.mparticle.com/work/[SQDSDKS-6551

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticle.java
+++ b/android-core/src/main/java/com/mparticle/MParticle.java
@@ -1269,18 +1269,18 @@ public class MParticle {
      * @see #logEvent(BaseEvent)
      */
     public enum EventType {
-        Unknown(0), Navigation(1), Location(2), Search(3), Transaction(4), UserContent(5), UserPreference(6), Social(7), Other(8), Media(9);
 
+        Unknown(0), Navigation(1), Location(2), Search(3), Transaction(4), UserContent(5), UserPreference(6),
+        Social(7), Other(8), Media(9), AddToCart(10), RemoveFromCart(11), Checkout(12), CheckoutOption(13),
+        Click(14), ViewDetail(15), Purchase(16), Refund(17), PromotionView(18), PromotionClick(19), AddToWishlist(20),
+        RemoveFromWishlist(21), Impression(22);
         private final int value;
 
         EventType(final int newValue) {
             value = newValue;
         }
 
-        public int getValue() {
-            return value;
-        }
-
+        public int getValue() { return value; }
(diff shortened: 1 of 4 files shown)


android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1269,18 +1269,18 @@ public static void removeListener(SdkListener listener) {
      * @see #logEvent(BaseEvent)
      */
     public enum EventType {
-        Unknown(0), Navigation(1), Location(2), Search(3), Transaction(4), UserContent(5), UserPreference(6), Social(7), Other(8), Media(9);
 
+        Unknown(0), Navigation(1), Location(2), Search(3), Transaction(4), UserContent(5), UserPreference(6),
```
samdozor, line 1273: this would be an incorrect public api change that'd likely cause a lot of pain and bugs :-) please revert

this enum is the one for custom events. it isn't meant to be used for all event types. ios does it this way and it's been a bug and complaint for years - customers for example log regular custom events with incorrect event types.

create some other private enum if we need it internally? these enum values shouldn't be exposed to customers.

android-kit-base/src/main/java/com/mparticle/kits/CommerceEventUtils.java
```diff
@@ -56,7 +56,42 @@ public static List<MPEvent> expandProductAction(CommerceEvent event) {
         List<Product> products = event.getProducts();
         if (products != null) {
             for (int i = 0; i < products.size(); i++) {
-                MPEvent.Builder itemEvent = new MPEvent.Builder(String.format(ITEM_NAME, productAction), MParticle.EventType.Transaction);
+                MPEvent.Builder itemEvent;
+                switch (productAction) {
+                    case Product.ADD_TO_CART:
+                        itemEvent = new MPEvent.Builder(String.format(ITEM_NAME, productAction), MParticle.EventType.AddToCart);
```
samdozor, line 62: and this is exactly what customers do and is not correct. then we end up with weird data, broken filtering, integrations, etc.

MPEvent is custom event and cannot have AddToCart (or any of the others below) as a custom event type in our schema.

samdozor (changes requested):
let's chat about another way to approach this, i'd like to understand the problem better.

PR #499, opened 2024-07-12
feat: Support direct URL routing

Mansi-mParticle:
## Instructions

 ## Summary
 - Cherry-pick commits from old branch https://github.com/mParticle/mparticle-android-sdk/pull/447 , routing data to endpoints.

 ## Testing Plan
 - Tested with sample application and executed unit test case

 - Closes https://go.mparticle.com/work/SQDSDKS-6550


android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -110,6 +111,11 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
             url = domainMappingUrl;
         }
         Uri uri;
+        Logger.debug("before Mansi -->"+url);
```
BrandonStalnaker, line 114: @Mansi-mParticle These two logs should be removed or the message should at least be change to something more generic and informative

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -110,6 +111,11 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
+        if (endpoint != Endpoint.CONFIG) {
+            url = NetworkUtilities.INSTANCE.getUrlWithPrefix(url, mConfigManager.getPodPrefix(), mConfigManager.isDirectUrlRoutingEnabled());
+            Logger.debug("Mansi -->"+url);
```
BrandonStalnaker, line 117: same as above

BrandonStalnaker (changes requested):
Th unit tests look good and I don't see any issues with the logic but I did find some logs that I think got leftover in testing.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1276,6 +1283,23 @@ public ConsentState getConsentState(long mpid) {
...
+        String prefix = "us1";
+        try {
+            String[] prefixFromApi = getApiKey().split("-");
+            if (prefixFromApi.length > 1) {
+                prefix = getApiKey().split("-")[0];
+            }
+        } catch (Exception e) {
+            prefix = "us1";
+        }
+        return prefix;
```
rmi22186, line 1300: I don't think that a try/catch is appropriate here.  try/catch should be used to code safely in case an unexpected exception is thrown.  In this case, we know there will or will not be a prefix. 
 It's not that the scenario without a prefix is an exception/error.  I think this should just be the if/else that you have on line 1294, but the `else` shoudl be defining the prefix as us1, as opposed to an error being caught.

android-core/src/main/java/com/mparticle/networking/NetworkOptionsManager.java
```diff
@@ -73,11 +74,11 @@ static String getDefaultUrl(Endpoint type) {
             case CONFIG:
                 return MPUtility.isEmpty(BuildConfig.MP_CONFIG_URL) ? MP_CONFIG_URL : BuildConfig.MP_CONFIG_URL;
             case IDENTITY:
-                return MPUtility.isEmpty(BuildConfig.MP_IDENTITY_URL) ? MP_IDENTITY_URL : BuildConfig.MP_IDENTITY_URL;
+                return MPUtility.isEmpty(BuildConfig.MP_IDENTITY_URL) ? MP_IDENTITY_URL_PREFIX : BuildConfig.MP_IDENTITY_URL;
```
rmi22186, line 77: we are now only returning the prefix?  where is the rest of the URL built?

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1276,6 +1283,23 @@ public ConsentState getConsentState(long mpid) {
...
+
+    public String getPodPrefix() {
+        String prefix = "us1";
+        try {
+            String[] prefixFromApi = getApiKey().split("-");
+            if (prefixFromApi.length > 1) {
+                prefix = getApiKey().split("-")[0];
+            }
+        } catch (Exception e) {
+            prefix = "us1";
```
rmi22186, line 1298: We should also always log the error whenever we catch an error.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
rmi22186: https://github.com/mParticle/mparticle-web-sdk/pull/801/files#diff-2717c7fb31c2070dd96f07394997809bc9be7f3fff03ab638db696cc480f39b6R491-R498

rmi22186: Account for the length being 0 in the if else clause.

android-core/src/main/kotlin/com.mparticle/NetworkUtilities.kt
```diff
@@ -0,0 +1,26 @@
...
+            val newUrl = if (enablePodRedirection) {
+                "$it.$podPrefix"
+            } else {
+                "$it"
+            }
+            "$newUrl.mparticle.com"
+        }
+    }
+
+    fun getPodPrefix(apiKey: String): String? {
```
rmi22186, line 19: let's move this to a test class since it is not being consumed in the core sdk at all

android-core/src/main/kotlin/com.mparticle/NetworkUtilities.kt
```diff
@@ -0,0 +1,26 @@
+package com.mparticle
+
+object NetworkUtilities {
```
rmi22186, line 3: let's get rid of this class and include the `getUrlWithPrefix` in the api client implementation

android-core/src/main/kotlin/com.mparticle/NetworkUtilities.kt
```diff
@@ -0,0 +1,26 @@
...
+    fun getUrlWithPrefix(
+        url: String?,
+        podPrefix: String,
+        enablePodRedirection: Boolean
+    ): String? {
+        return url?.let {
+            val newUrl = if (enablePodRedirection) {
```
rmi22186, line 10: Refactor this function as we discussed on the call

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -110,6 +111,9 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
             url = domainMappingUrl;
         }
         Uri uri;
+        if (endpoint != Endpoint.CONFIG) {
+            url = NetworkUtilities.INSTANCE.getUrlWithPrefix(url, mConfigManager.getPodPrefix(), mConfigManager.isDirectUrlRoutingEnabled());
```
rmi22186, line 115: should get the url with prefix ONLY if the boolean is true here.

rmi22186 (changes requested):
Looks and works great overall.  I think there are some areas that can be refactored to be more understandable.

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -20,11 +21,150 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
...
+        val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
+        val url =
+            baseClientImpl.getPodUrl(NetworkOptionsManager.MP_URL_PREFIX, prefix, false)
+        assertEquals("${NetworkOptionsManager.MP_URL_PREFIX}.mparticle.com", url)
+    }
+
+    @Test
+    fun testAllPrefixes() {
+        val map = mapOf<String, String>(
+            Pair("us1-1vc4gbp24cdtx6e31s58icnymzy83f1uf", "us1"),
```
rmi22186, line 100: just so future devs know, add a comment above here that says these are fake api keys.

i would also test one where there is no prefix, so that legacy us1 is covered

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1276,6 +1283,28 @@ public ConsentState getConsentState(long mpid) {
...
 
+    public boolean isDirectUrlRoutingEnabled() {
+        return directUrlRouting;
+    }
+
+    /* When Direct URL Routing is true, we create a new set of baseUrls that include the silo in the urls. mParticle API keys are prefixed with the
+     silo and a hyphen (ex. "us1-", "us2-", "eu1-").  us1 was the first silo,and before other silos existed, there were no prefixes and all apiKeys
+     were us1. As such, if we split on a '-' and the resulting array length is 1, then it is an older APIkey that should route to us1.
+     When splitKey.length is greater than 1, then splitKey[0] will be us1, us2, eu1, au1, or st1, etc as new silos are added*/
+    public String getPodPrefix() {
```
rmi22186, line 1294: let's use this instead of the getPodPrefix in the test class.  this way we will be adding unit test coverage to this method

rmi22186: or copy and paste this logic to the test if this is not possible

rmi22186 (changes requested):

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -406,6 +408,11 @@ private synchronized void updateCoreConfig(JSONObject responseJSON, boolean newC
             editor.putString(KEY_PUSH_MESSAGES, sPushKeys.toString());
         }
 
+        if (responseJSON.has(KEY_DIRECT_URL_ROUTING)) {
+           directUrlRouting = responseJSON.optBoolean(KEY_DIRECT_URL_ROUTING);
+            editor.putBoolean(KEY_DIRECT_URL_ROUTING, directUrlRouting);
+        }
+
```
rmi22186, line 415: Small formatting nit

```suggestion
        if (responseJSON.has(KEY_DIRECT_URL_ROUTING)) {
           directUrlRouting = responseJSON.optBoolean(KEY_DIRECT_URL_ROUTING);
           editor.putBoolean(KEY_DIRECT_URL_ROUTING, directUrlRouting);
        }

```

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -20,11 +23,166 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
...
+            .build()
+        MParticle.start(options)
+        val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
+        map.forEach { key, value ->
+            mConfigManager?.setCredentials(key, value)
+            val prefix = mConfigManager?.podPrefix
+            assertEquals(value, prefix)
+            assertEquals(
+                "${NetworkOptionsManager.MP_URL_PREFIX}.$prefix.mparticle.com",
+                baseClientImpl.getPodUrl(NetworkOptionsManager.MP_URL_PREFIX, prefix, true)
```
rmi22186, line 156: nit:  

Can you update the formatting of this assert to match the formatting of lines 160-164, 168-172, and 176-180?

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -20,11 +23,166 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
...
+                    prefix,
+                    false
+                )
+            )
+        }
+    }
+
     @Test
     @Throws(MalformedURLException::class)
     fun testGetUrlForceDefaultOption() {
```
rmi22186, line 188: This test would be removed in the future since we are trying to get rid of that `forceDefaultUrl` boolean since that was only for the recursion issue, right?

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -56,17 +214,112 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
...
-            Assert.assertEquals(defaultUrls[endpoint].toString(), generatedUrl.toString())
-            TestCase.assertTrue(generatedUrl === generatedUrl.defaultUrl)
+            assertEquals(defaultUrls[endpoint].toString(), generatedUrl.defaultUrl.toString())
         }
         for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
             val generatedUrl = baseClientImpl.getUrl(endpoint, endpoint.name, false)
             Assert.assertNotEquals(defaultUrls[endpoint].toString(), generatedUrl.toString())
             Assert.assertFalse(generatedUrl === generatedUrl.defaultUrl)
-            Assert.assertEquals(
+            assertEquals(
```
einsteinx2, line 223: Is there a reason that some checks use `Assert.assert...` and others just use `assert...`? In this case the existing `Assert.assertEquals` was even changed to remove the `Assert.` but the other asserts still have it?

android-core/src/androidTest/kotlin/com.mparticle/networking/NetworkOptionsTest.kt
```diff
@@ -181,6 +188,161 @@ class NetworkOptionsTest : BaseCleanInstallEachTest() {
...
+            NetworkOptionsManager.MP_URL_PREFIX.addSuffix(),
+            mpClient.getUrl(MParticleBaseClientImpl.Endpoint.EVENTS).authority
+        )
+        Assert.assertEquals(
+            NetworkOptionsManager.MP_IDENTITY_URL_PREFIX.addSuffix(),
+            mpClient.getUrl(MParticleBaseClientImpl.Endpoint.IDENTITY).authority
+        )
+    }
+
+    /*
```
einsteinx2, line 314: If this test can't run, should we just remove it? I don't like having big commented out blocks of code in the codebase... We can always pull it back out of git history if/when we're ready to add it back.

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
```diff
@@ -1276,6 +1283,28 @@ public ConsentState getConsentState(long mpid) {
...
+    }
+
+    /* When Direct URL Routing is true, we create a new set of baseUrls that include the silo in the urls. mParticle API keys are prefixed with the
+     silo and a hyphen (ex. "us1-", "us2-", "eu1-", etc).  us1 was the first silo,and before other silos existed, there were no prefixes and all apiKeys
+     were us1. As such, if we split on a '-' and the resulting array length is 1, then it is an older APIkey that should route to us1.
+     When splitKey.length is greater than 1, then splitKey[0] will be us1, us2, eu1, au1, or st1, etc as new silos are added*/
+    public String getPodPrefix() {
+        String prefix = "us1";
+        try {
+            String[] prefixFromApi = getApiKey().split("-");
```
einsteinx2, line 1297: Will this throw if there's no `-` in the api key or will it just result in `prefixFromApi.length == 1`? If so, what are we catching here?

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -133,44 +147,86 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
+            }
+            if (path != null && !path.isEmpty()) {
+                uriBuilder.path(path);
+            }
+            return MPUrl.getUrl(uriBuilder.build().toString(), null);
+        }
+        return null;
+    }
+
+    String getPodUrl(
```
einsteinx2, line 218: Is there a reason for this method definition to be split on 5 lines? It seems like the definition on line 204 is longer but is only one line? I know we have a linter that runs, but this seems inconsistent and in general, at least personally, I find this type of formatting harder to read than more compact formatting but that may just be my preference...

android-core/src/main/java/com/mparticle/internal/ConfigManager.java
Mansi-mParticle: If there is no '-' in the string, it means the prefix length from the API is 1. The catch block handles cases where the field might be null or empty, ensuring the SDK doesn’t crash. (This is useful for reset method.)

android-core/src/androidTest/kotlin/com.mparticle/networking/NetworkOptionsTest.kt
rmi22186: Mansi - can you make a story ticket for this and add it under [this epice](https://mparticle-eng.atlassian.net/browse/SQDSDKS-6612)?  You could copy and paste the below code in there to easily bring it back in a future follow up.  I think the ticket should be to add the library you were talking about but said you were uncertain if you could add in order to mock the request.

Mansi-mParticle: @rmi22186  created here SQDSDKS-6613

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
Mansi-mParticle: yes agree in future will remove this test cases

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
Mansi-mParticle: No, there’s no difference. When I rewrote this, the Assert class was automatically imported, so we can use assertEquals directly.

einsteinx2 (approved):
lgtm

rmi22186 (approved):
Approved, but let's wait for Sam's confirmation about the build.gradle URL files and its relationship to this feature.

android-core/src/androidTest/kotlin/com.mparticle/networking/NetworkOptionsTest.kt
```diff
@@ -181,6 +188,130 @@ class NetworkOptionsTest : BaseCleanInstallEachTest() {
         )
     }
 
+    // TODO : https://go.mparticle.com/work/SQDSDKS-6550
```
rmi22186, line 191:
```suggestion
    // https://go.mparticle.com/work/SQDSDKS-6550
```

rmi22186 (approved):

samdozor (approved):

BrandonStalnaker (approved):

PR #501, opened 2024-08-02
fix: Remove recursion from the method and add default URL generation.

Mansi-mParticle:
## Instructions

 ## Summary
 - Remove recursion from the method and add a feature to create a default URL. This includes a default URL for HMAC signatures when domain mapping is used.

 ## Testing Plan
 - Tested with sample application and executed unit test case

 - Closes https://go.mparticle.com/work/SQDSDKS-6550

### DIFF
--- a/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
+++ b/android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
@@ -1,11 +1,12 @@
 package com.mparticle.networking
 
+import android.net.Uri
+import com.mparticle.BuildConfig
 import com.mparticle.MParticle
 import com.mparticle.MParticleOptions
 import com.mparticle.internal.AccessUtils
 import com.mparticle.internal.ConfigManager
 import com.mparticle.testutils.BaseCleanInstallEachTest
-import junit.framework.TestCase
 import org.junit.Assert
 import org.junit.Assert.assertEquals
 import org.junit.Before
@@ -213,8 +214,7 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
         val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
         for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
             val generatedUrl = baseClientImpl.getUrl(endpoint, endpoint.name, true)
-            assertEquals(defaultUrls[endpoint].toString(), generatedUrl.toString())
-            TestCase.assertTrue(generatedUrl === generatedUrl.defaultUrl)
+            assertEquals(defaultUrls[endpoint].toString(), generatedUrl.defaultUrl.toString())
         }
         for (endpoint in MParticleBaseClientImpl.Endpoint.values()) {
             val generatedUrl = baseClientImpl.getUrl(endpoint, endpoint.name, false)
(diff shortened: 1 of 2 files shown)


android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -103,26 +103,34 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
         NetworkOptions networkOptions = mConfigManager.getNetworkOptions();
         DomainMapping domainMapping = networkOptions.getDomain(endpoint);
         String url = NetworkOptionsManager.getDefaultUrl(endpoint);
+        // Default domain to use for URL generation when domain mapping is specified
+        String defaultDomain = url ;
```
rmi22186, line 107:
```suggestion
        String defaultDomain = url;
```

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -103,26 +103,34 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
         boolean isDefaultUrl = true;
-        if (domainMapping != null && !MPUtility.isEmpty(domainMapping.getUrl()) && !forceDefaultUrl) {
+        if (domainMapping != null && !MPUtility.isEmpty(domainMapping.getUrl())) {
```
rmi22186, line 109: if `forceDefaultUrl` is removed from here, we should remove it from the function signature as well

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -103,26 +103,34 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
         NetworkOptions networkOptions = mConfigManager.getNetworkOptions();
         DomainMapping domainMapping = networkOptions.getDomain(endpoint);
         String url = NetworkOptionsManager.getDefaultUrl(endpoint);
+        // Default domain to use for URL generation when domain mapping is specified
```
rmi22186, line 106: Clarify what to use when it is not specified here also

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -103,26 +103,34 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
+        if (domainMapping != null && !MPUtility.isEmpty(domainMapping.getUrl())) {
             String domainMappingUrl = domainMapping.getUrl();
             isDefaultUrl = url.equals(domainMappingUrl);
             url = domainMappingUrl;
         }
         Uri uri;
-        if (endpoint != Endpoint.CONFIG & isDefaultUrl) {
-            url = getPodUrl(url, mConfigManager.getPodPrefix(), mConfigManager.isDirectUrlRoutingEnabled());
+        if (endpoint != Endpoint.CONFIG) {
+            if (isDefaultUrl) {
```
rmi22186, line 116: this means that the domain mapping is not set right?  I would add that as a comment here.  This code gets super complicated, so the more comments the better.

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
Mansi-mParticle: As discussed, we will not remove it at this time, as it would require significant changes to the getURL method.

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -136,43 +145,71 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
                         .build();
-                return MPUrl.getUrl(uri.toString(), defaultUrl);
+                return MPUrl.getUrl(uri.toString(), !isDefaultUrl ? generateDefaultURL(uri, defaultDomain, (SERVICE_VERSION_1 + "/identity/" + pathPostfix)) : null);
             case IDENTITY:
                 subdirectory = overridesSubdirectory ? "" : SERVICE_VERSION_1 + "/";
+                pathPostfix = identityPath;
                 uri = new Uri.Builder()
                         .scheme(BuildConfig.SCHEME)
                         .encodedAuthority(url)
                         .path(subdirectory + identityPath)
```
rmi22186, line 174: to be consistent with everywhere else, maybe this should be 
```suggestion
                        .path(subdirectory + pastPostfix)
```

android-core/src/main/java/com/mparticle/networking/MParticleBaseClientImpl.java
```diff
@@ -136,43 +145,71 @@ protected MPUrl getUrl(Endpoint endpoint, @Nullable String identityPath, boolean
...
                         .build();
-                return MPUrl.getUrl(uri.toString(), defaultUrl);
+                return MPUrl.getUrl(uri.toString(), !isDefaultUrl ? generateDefaultURL(uri, defaultDomain, (SERVICE_VERSION_1 + "/" + pathPostfix)) : null);
             case AUDIENCE:
+                pathPostfix = SERVICE_VERSION_2 + "/" + mApiKey + "/audience?mpID=" + mConfigManager.getMpid();
                 uri = new Uri.Builder()
                         .scheme(BuildConfig.SCHEME)
                         .encodedAuthority(url)
-                        .path(SERVICE_VERSION_2 + "/" + mApiKey + "/audience?mpID=" + mConfigManager.getMpid())
+                        .path(pathPostfix)
```
rmi22186, line 182: to be consistent with the other cases, you should do:
```
subdirectory = SERVICE_VERSION_2 + "/"
pathPostfix =mApiKey + "/audience?mpID=" + mConfigManager.getMpid()
```
then 


```
.path(subdirectory + pathPostfix)
```

rmi22186 (changes requested):
minor nits

android-core/src/androidTest/kotlin/com.mparticle/networking/MParticleBaseClientImplTest.kt
```diff
@@ -226,4 +226,100 @@ class MParticleBaseClientImplTest : BaseCleanInstallEachTest() {
...
+            .build()
+        MParticle.start(options)
+        val baseClientImpl = AccessUtils.getApiClient() as MParticleBaseClientImpl
+        val subdirectory = "/v2/"
+        val uri = Uri.Builder()
+            .scheme(BuildConfig.SCHEME)
+            .encodedAuthority("nativesdks.us1.mparticle.com")
+            .path("$subdirectory fooApis/events")
+            .build()
+        val result = baseClientImpl.generateDefaultURL(uri, "nativesdks.mparticle.com", "v2/us1-bee5781b649a7a40a592c2000bc892d0/events")
```
rmi22186, line 244: please use a fake API Key, like `us1-foo`.  comment holds true in all tests

rmi22186 (approved):

rmi22186:
I would do a squash and merge into a single commit into the other feature branch.

PR #502, opened 2024-08-05
fix: Remove Code for Handling Google AdId on a Separate Thread

Mansi-mParticle:
## Instructions

 ## Summary
 - Move the code for retrieving the Google AD ID from a separate thread to the main thread. On some OS, running it in a separate thread causes the main thread to be delayed. (Revert Changes from PR #491 )
 

 ## Testing Plan
 - Test with a sample app on Android os 14 and 15 using the simulator.

 - Closes https://go.mparticle.com/work/SQDSDKS-6594


einsteinx2 (approved):
Confirmed this reverts the previous PR #491 and agree with the proposed solution to revert.

PR #503, opened 2024-08-09
fix: Crash in KitConfiguration due to NumberFormatException

Mansi-mParticle:
## Instructions

 ## Summary
 - Fix rare crash in KitConfiguration caused by NumberFormatException. Updated code to handle NumberFormatException and null cases.

 ## Testing Plan
 - Tested with the sample application and could not reproduce the issue initially. However, after adding a unit test, I was able to reproduce the issue and address it.

 - Closes https://go.mparticle.com/work/SQDSDKS-6548

### DIFF
--- a/android-kit-base/src/test/kotlin/com/mparticle/kits/KitConfigurationTest.kt
+++ b/android-kit-base/src/test/kotlin/com/mparticle/kits/KitConfigurationTest.kt
@@ -722,6 +722,15 @@ class KitConfigurationTest {
     }
 
     @Test
+    fun testConvertToSparseArray_When_JSON_IS_NULL() {
+        val kitConfiguration = MockKitConfiguration()
+        val method: Method = MockKitConfiguration::class.java.getDeclaredMethod("convertToSparseArray", JSONObject::class.java)
+        method.isAccessible = true
+        val result = method.invoke(kitConfiguration, null) as SparseBooleanArray
+        Assert.assertEquals(0, result.size())
+    }
+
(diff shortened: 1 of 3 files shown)


android-kit-base/src/test/kotlin/com/mparticle/kits/KitConfigurationTest.kt
```diff
@@ -676,6 +678,97 @@ class KitConfigurationTest {
...
+        """.trimIndent()
+        val jsonConfiguration = JSONObject(jsonData)
+        val method: Method = MockKitConfiguration::class.java.getDeclaredMethod("convertToSparseArray", JSONObject::class.java)
+        method.isAccessible = true
+        val result = method.invoke(kitConfiguration, jsonConfiguration) as SparseBooleanArray
+        Assert.assertEquals(12, result.size())
+    }
+
+    @Test
+    fun testConvertToSparseArray_When_JSON_OBJECT_IS_NULL() {
```
rmi22186, line 708: Isn't this JSON object just an empty object as opposed to a `null`?  Should there be a test case for 
`{"ec" : null}` or no?

Mansi-mParticle: Add unit test case for null validation

rmi22186 (approved):

einsteinx2 (approved):
LGTM, reviewed live on Zoom

PR #504, opened 2024-08-21
feat: Certificate pinning update

Mansi-mParticle:
## Instructions

 ## Summary
 - Update Root Certificates

 ## Testing Plan
 - Tested with a sample app and used domain mapping to test Let's Encrypt certificates.

 - Closes https://go.mparticle.com/work/SQDSDKS-6546


einsteinx2 (approved):
Reviewed approach and test cases with Mansi.

PR #511, opened 2024-10-01
refactor: Migrate Internal Application context class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s ApplicationContextWrapper,AppStateManager classes to Kotlin.
 - Updated Deprecated method.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6706


android-core/src/main/kotlin/com/mparticle/internal/ApplicationContextWrapper.kt
```diff
@@ -0,0 +1,351 @@
...
+        get() = mRecord
+        set(record) {
+            if (record.also { this.mRecord = it }) {
+                startRecordLifecycles()
+            } else {
+                stopRecordLifecycles()
+            }
+        }
+
+    @SuppressLint("MissingSuperCall")
```
einsteinx2, line 48: Do we truly not want to be calling super here and in the methods below?

android-core/src/main/kotlin/com/mparticle/internal/ApplicationContextWrapper.kt
```diff
@@ -0,0 +1,351 @@
...
+        bundle: Bundle?
+    ): LifeCycleEvent {
+        return LifeCycleEvent(methodType, activityRef, bundle)
+    }
+
+    @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+    inner class ActivityLifecycleCallbackRecorder : ActivityLifecycleCallbacks {
+        var lifeCycleEvents: MutableList<LifeCycleEvent> =
+            Collections.synchronizedList(LinkedList())
+        var MAX_LIST_SIZE: Int = 10
```
einsteinx2, line 170:
```suggestion
        val MAX_LIST_SIZE: Int = 10
```
This should be a constant

android-core/src/main/kotlin/com/mparticle/internal/ApplicationContextWrapper.kt
```diff
@@ -0,0 +1,351 @@
...
+                        while (recordedLifecycleList.size > 0) {
+                            val lifeCycleEvent = recordedLifecycleList.removeFirst()
+                            if (lifeCycleEvent.activityRef != null) {
+                                val recordedActivity = lifeCycleEvent.activityRef.get()
+                                if (recordedActivity != null) {
+                                    if (recordedActivity === currentActivity) {
+                                        when (lifeCycleEvent.methodType) {
+                                            MethodType.ON_CREATED -> {
+                                                Logger.debug("Forwarding OnCreate")
+                                                callback!!.onActivityCreated(
```
einsteinx2, line 303: Is there any way we can write this without force unwrapping? Same for all uses of `!!` below.

android-core/src/androidTest/kotlin/com.mparticle/SessionMessagesTest.kt
```diff
@@ -28,10 +28,10 @@ class SessionMessagesTest : BaseCleanStartedEachTest() {
     fun testSessionStartMessage() {
         val sessionStartReceived = BooleanArray(1)
         sessionStartReceived[0] = false
-        Assert.assertFalse(mAppStateManager.session.isActive)
+        Assert.assertFalse(mAppStateManager.fetchSession().isActive)
```
einsteinx2, line 31: Why did this change to `fetchSession()`? I don't see that in the code changes.

einsteinx2 (changes requested):

android-core/src/androidTest/kotlin/com.mparticle/SessionMessagesTest.kt
Mansi-mParticle: so with same name it's kotlin compiler issue, so It was giving compiling error for conflict . Hence I changed the method name.

android-core/src/main/kotlin/com/mparticle/internal/ApplicationContextWrapper.kt
Mansi-mParticle: yes it's required here because we need here our custom logic so.

einsteinx2 (approved):

android-core/src/main/kotlin/com/mparticle/internal/ApplicationContextWrapper.kt
```diff
@@ -0,0 +1,351 @@
...
+            return false
+        }
+    }
+
+    internal inner class ReplayLifecycleCallbacksRunnable(var callback: ActivityLifecycleCallbacks) :
+        Runnable {
+        @TargetApi(Build.VERSION_CODES.ICE_CREAM_SANDWICH)
+        override fun run() {
+            if (callback != null && mActivityLifecycleCallbackRecorder != null && isReplayActivityLifecycle) {
+                val reference = if (MParticle.getInstance()!!
```
einsteinx2, line 281: I'm seeing the non-null assertion operator used all over the place here...

We can not use this in production code ever, unless it's absolutely necessary for some reason and immediately preceeded by a null check or the variable declaration or assignment of a non-null value (and even then we should not use it except as a last resort).

Please see the Kotlin docs here: https://kotlinlang.org/docs/null-safety.html#not-null-assertion-operator

einsteinx2 (changes requested):

android-core/src/main/kotlin/com/mparticle/internal/AppStateManager.kt
```diff
@@ -0,0 +1,494 @@
...
+        } catch (e: Exception) {
+            Logger.verbose("Failed while trying to track activity pause: " + e.message)
+        }
+    }
+
+    fun ensureActiveSession() {
+        if (!mInitialized) {
+            initialize(null, null, null, null)
+        }
+        val session = session
```
einsteinx2, line 241:
```suggestion
```

android-core/src/main/kotlin/com/mparticle/internal/AppStateManager.kt
```diff
@@ -0,0 +1,494 @@
...
+            val minDistance = mPreferences.getLong(Constants.PrefKeys.LOCATION_MINDISTANCE, 0)
+            if (provider != null && minTime > 0 && minDistance > 0) {
+                val instance = MParticle.getInstance()
+                instance?.enableLocationTracking(provider, minTime, minDistance)
+            }
+        }
+    }
+
+    fun shouldEndSession(): Boolean {
+        val session = session
```
einsteinx2, line 304:
```suggestion
```

einsteinx2 (changes requested):

einsteinx2 (approved):

einsteinx2 (approved):

PR #513, opened 2024-10-08
feat: Identity API response caching

Mansi-mParticle:
## Instructions

 ## Summary
 - Adds caching of identify and login calls and Clears that cache on modify and logout calls.
 
 ## Testing Plan
 - Tested with sample application. And currently Executed below Test cases:
1)Login with same user on same session and on cold launch.
2)Login with two different user on same session and on cold launch.
3)Call Identity with same details as login details.
4)Call Identity with same user details on same session and on cold launch.

 - Closes https://go.mparticle.com/work/SQDSDKS-6673


android-core/src/androidTest/kotlin/com.mparticle/identity/IdentityApiTest.kt
```diff
@@ -111,26 +113,27 @@ class IdentityApiTest : BaseCleanStartedEachTest() {
 
         // test that change actually took place
         result.addSuccessListener { identityApiResult ->
-            Assert.assertEquals(identityApiResult.user.id, mpid1)
-            Assert.assertEquals(identityApiResult.previousUser!!.id, mStartingMpid.toLong())
+            Assert.assertEquals(identityApiResult.user.id, mStartingMpid)
+            // After Adding Identity caching, it uses previous response
+            // Assert.assertEquals(identityApiResult.previousUser!!.id, mStartingMpid.toLong())
```
einsteinx2, line 118: If we don't need these asserts, can we remove hem instead of commenting them out? Same for the other 2 in this file.

einsteinx2 (changes requested):
Other than the commented out asserts, the added tests (SQDSDKS-6674) looks good to me

android-core/src/main/java/com/mparticle/identity/IdentityApi.java
```diff
@@ -348,6 +350,26 @@ private void reset() {
...
+            // Convert byte array to hex string
+            StringBuilder hexString = new StringBuilder();
+            for (byte b : hashBytes) {
+                String hex = Integer.toHexString(0xff & b);
+                if (hex.length() == 1) hexString.append('0');
+                hexString.append(hex);
+            }
+            return hexString.toString();
+        } catch (NoSuchAlgorithmException e) {
+            //throw new RuntimeException("Hashing algorithm not found", e);
```
einsteinx2, line 367: This feels like an anti-pattern... What do we do if the device doesn't have SHA-256 available? Can this be rewritten to use fallback hashes like SHA-1 or others?

android-core/src/main/java/com/mparticle/identity/IdentityApiRequest.java
```diff
@@ -211,4 +214,47 @@ public Builder userAliasHandler(@Nullable UserAliasHandler userAliasHandler) {
...
+            byte[] hashBytes = md.digest(input.getBytes());
+            StringBuilder hexString = new StringBuilder();
+            for (byte b : hashBytes) {
+                String hex = Integer.toHexString(0xff & b);
+                if (hex.length() == 1) hexString.append('0');
+                hexString.append(hex);
+            }
+            return hexString.toString().substring(0, 16); // Shorten to first 16 characters
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException(e);
```
einsteinx2, line 257: Here I see we throw the exception instead of swallowing it. Can we also have a fallback here? Also this is duplicating the same code as the `generateHash` method above, can we move that to a utility class and reuse it instead?

android-core/src/main/java/com/mparticle/identity/MParticleIdentityClientImpl.java
```diff
@@ -75,18 +82,33 @@ public MParticleIdentityClientImpl(Context context, ConfigManager configManager,
 
     public IdentityHttpResponse login(IdentityApiRequest request) throws JSONException, IOException {
         JSONObject jsonObject = getStateJson(request);
+        IdentityHttpResponse existsResponse = checkIfExists(request, LOGIN_CALL);
+
+        if (existsResponse != null) {
+            return existsResponse;
+        }
+        Long maxAgeTime=86400L;
```
einsteinx2, line 90: Would this be better as a constant in the class instead of defined in each method?

einsteinx2 (changes requested):

android-core/src/main/java/com/mparticle/identity/MParticleIdentityClientImpl.java
```diff
@@ -129,6 +165,57 @@ public IdentityHttpResponse modify(IdentityApiRequest request) throws JSONExcept
...
+                    return null;
+
+                }
+            } catch (Exception e) {
+                Logger.error("Exception  " + e);
+
+            }
+        }
+        return null;
+    }
```
einsteinx2, line 217:
```suggestion
    private IdentityHttpResponse checkIfExists(IdentityApiRequest request, String callType) {
        if (!mConfigManager.isIdentityCacheFlagEnabled()) {
            return null;
        }

        try {
            String key = identityCacheKey(request, callType);
            if (key == null) {
                return null;
            }

            if (identityCacheMap.isEmpty()) {
                identityCacheMap = mConfigManager.fetchIdentityCache();
            }

            IdentityHttpResponse cachedResponse = identityCacheMap.get(key);
            if (cachedResponse != null) {
                if (cachedResponse.getCacheExpirationMillis() > System.currentTimeMillis()) {
                    return cachedResponse;
                } else {
                    // Expired, so remove from cache
                    mConfigManager.removeFromIdentityCache(key);
                }
            }
        } catch (Exception e) {
            Logger.error("Exception  " + e);
        }
        return null;
    }
```

Using the suggestion I made in the `ConfigManager` class, this method would look something like the above.

einsteinx2 (changes requested):

thomson-t:
Closing as outdated. This work has been ported onto latest main in https://github.com/mParticle/mparticle-android-sdk/pull/786.

PR #516, opened 2024-10-25
refactor: Migrate Internal user storage class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s Migrate UserStorage class to kotlin

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6721


android-core/src/main/kotlin/com/mparticle/internal/UserStorage.kt
```diff
@@ -0,0 +1,604 @@
...
+                    FIRST_SEEN_TIME, messageManagerSharedPreferences.getLong(
+                        Constants.PrefKeys.INSTALL_TIME, defaultSeenTime
+                    )
+                ).apply()
+            }
+            return mPreferences.getLong(FIRST_SEEN_TIME, defaultSeenTime)
+        }
+        set(time) {
+            if (!mPreferences.contains(FIRST_SEEN_TIME)) {
+                mPreferences.edit().putLong(FIRST_SEEN_TIME, time!!).apply()
```
einsteinx2, line 224: Looks like we're force unwrapping time without checking for null, and we explicitly allow null since this is marked as `Long?`. So this will throw a null pointer exception whenever `time` is null...

We should never be using `!!` anywhere in production code, only in tests.

android-core/src/main/kotlin/com/mparticle/internal/UserStorage.kt
```diff
@@ -0,0 +1,604 @@
...
+
+    var lastSeenTime: Long?
+        get() {
+            if (!mPreferences.contains(LAST_SEEN_TIME)) {
+                mPreferences.edit().putLong(LAST_SEEN_TIME, defaultSeenTime).apply()
+            }
+            return mPreferences.getLong(LAST_SEEN_TIME, defaultSeenTime)
+        }
+        set(time) {
+            mPreferences.edit().putLong(LAST_SEEN_TIME, time!!).apply()
```
einsteinx2, line 236: Same here with the `!!`

android-core/src/main/kotlin/com/mparticle/internal/UserStorage.kt
```diff
@@ -0,0 +1,604 @@
...
+                const val TOTAL_SINCE_UPGRADE: String = "mp::launch_since_upgrade"
+                const val USER_IDENTITIES: String = "mp::user_ids::"
+            }
+        }
+
+        fun migrate(userStorage: UserStorage) {
+            try {
+                userStorage.deletedUserAttributes = getDeletedUserAttributes
+                userStorage.previousSessionId = previousSessionId
+                val ltv: String = ltv!!
```
einsteinx2, line 361: Same here with the !!

android-core/src/main/kotlin/com/mparticle/internal/UserStorage.kt
```diff
@@ -0,0 +1,604 @@
...
+                if (previousSessionStart != 0L) {
+                    userStorage.setPreviousSessionStart(previousSessionStart)
+                }
+                val totalRuns: Int = totalRuns
+                if (totalRuns != 0) {
+                    userStorage.setTotalRuns(totalRuns)
+                }
+
+                //migrate both cookies and device application stamp
+                val cookies: String = cookies!!
```
einsteinx2, line 391: Same here with the !!

einsteinx2 (changes requested):

einsteinx2 (approved):

PR #518, opened 2024-10-31
fix: previous Google ad Id from being null on cold launch

Mansi-mParticle:
## Instructions

 ## Summary
 - Fix a bug where the previous Google app ID is null when the app is relaunched. This happened because the app ID was stored when the app paused; now, the code has been moved to when the app is in the foreground.

 ## Testing Plan
 - Tested with sample app and Executed unit test cases

 - Closes https://go.mparticle.com/work/SQDSDKS-6830

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/MPUtility.java
+++ b/android-core/src/main/java/com/mparticle/internal/MPUtility.java
@@ -70,6 +70,7 @@ public class MPUtility {
     private static String sOpenUDID;
     private static final char[] HEX_CHARS = "0123456789abcdef".toCharArray();
     private static final String TAG = MPUtility.class.toString();
+    private static AdIdInfo infoId = null;
 
     public static long getAvailableMemory(Context context) {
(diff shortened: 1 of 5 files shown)


android-core/src/main/java/com/mparticle/internal/MPUtility.java
```diff
@@ -70,6 +70,7 @@ public class MPUtility {
     private static String sOpenUDID;
     private static final char[] HEX_CHARS = "0123456789abcdef".toCharArray();
     private static final String TAG = MPUtility.class.toString();
+    private static AdIdInfo infoId = null;
```
einsteinx2, line 73:
```suggestion
    private static AdIdInfo adInfoId = null;
```
A nitpick, but maybe this should be named `adInfoId` for clarity as `infoId` is fairly generic and this utility class does various things.

einsteinx2 (changes requested):

einsteinx2 (approved):
lgtm

PR #520, opened 2024-11-08
refactor: Migrate Internal listeners class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
- Migrate internal listeners package classes to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6704


android-core/src/main/kotlin/com/mparticle/internal/listeners/InternalListener.kt
```diff
@@ -84,7 +78,12 @@ public interface InternalListener {
      * @param body    the request body
      * @param objects any underlying objects that the request body is derived from, for example, an IdentityApiRequest instance
      */
-    void onNetworkRequestStarted(SdkListener.Endpoint type, String url, JSONObject body, Object... objects);
+    fun onNetworkRequestStarted(
+        type: SdkListener.Endpoint?,
+        url: String?,
```
einsteinx2, line 83: Are these supposed to be nullable?

android-core/src/main/kotlin/com/mparticle/internal/listeners/InternalListener.kt
```diff
@@ -93,30 +92,35 @@ public interface InternalListener {
      * @param response     the response body
      * @param responseCode the response code
      */
-    void onNetworkRequestFinished(SdkListener.Endpoint type, String url, JSONObject response, int responseCode);
+    fun onNetworkRequestFinished(
+        type: SdkListener.Endpoint,
+        url: String,
```
einsteinx2, line 97: For example here they are not nullable

android-core/src/main/kotlin/com/mparticle/internal/listeners/InternalListener.kt
```diff
@@ -93,30 +92,35 @@ public interface InternalListener {
...
+    )
 
     /**
      * this should be called when the current Session changes, for example, it starts, stops or the
      * event count changes
      *
      * @param internalSession
      */
-    void onSessionUpdated(InternalSession internalSession);
+    fun onSessionUpdated(internalSession: InternalSession?)
```
einsteinx2, line 108: Can internalSession every be null?

einsteinx2 (approved):

PR #521, opened 2024-11-19
refactor: Migrate Internal messages package classes to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal messages package classes BaseMPMessage and BaseMPMessageBuilder class and  to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6705


android-core/src/main/kotlin/com/mparticle/internal/messages/BaseMPMessage.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.json.JSONException
+import org.json.JSONObject
+
+open class BaseMPMessage : JSONObject {
+    var mpId: Long = 0
+        private set
+
+    protected constructor()
+
+    constructor(builder: BaseMPMessageBuilder, session: InternalSession, location: Location?, mpId: Long) : super(builder, builder.keys) {
```
einsteinx2, line 16: I see in the original file this was explicitly marked `protected`, is that not necessary?

android-core/src/main/kotlin/com/mparticle/internal/messages/BaseMPMessage.kt
```diff
@@ -0,0 +1,74 @@
...
+        get() {
+            try {
+                return getLong(Constants.MessageKey.TIMESTAMP)
+            } catch (_: JSONException) {
+            }
+            return 0
+        }
+
+    val sessionId: String
+        get() = if (Constants.MessageType.SESSION_START == messageType) {
```
einsteinx2, line 57: From a style perspective, should we standardize on the format used in line 48 (i.e. no `=` and using `{}` for any getting longer than one line? I know this technically works because the if/else is treated as a single expression, but it seems like it's cleaner/easier to read if we use {} around any multiline getters.

einsteinx2: I'm ok with using the `get() = ` format for single line getters

android-core/src/main/kotlin/com/mparticle/internal/messages/BaseMPMessageBuilder.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    val keys: Array<String?>
+        get() {
+            val strings: MutableList<String?> = ArrayList<String?>()
+            val iterator = keys()
+            while (iterator.hasNext()) {
+                strings.add(iterator.next())
+            }
+            return strings.toTypedArray<String?>()
+        }
```
einsteinx2, line 131: We should try to use functional style in cases like this as it's much more concise (please confirm the syntax in my suggestion actually works correctly and double check performance before committing the change, but if performance is similar, I think the functional style is much cleaner and easier to read.

```suggestion
    val keys: Array<String?>
        get() = keys().asSequence().toList().toTypedArray()
```

einsteinx2 (changes requested):

android-core/src/main/kotlin/com/mparticle/internal/messages/BaseMPMessage.kt
Mansi-mParticle: Kotlin: A protected constructor is not accessible to other classes in the same package, only in subclasses. and we are using in same package class BaseMPMessageBuilder.

einsteinx2 (approved):

PR #524, opened 2024-11-29
refactor: Migrate Internal DeviceAttributes class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
- Migrate internal package’s DeviceAttributes classe to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6711


android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+    private var deviceInfo: JSONObject? = null
+    private var appInfo: JSONObject? = null
+    private var firstCollection = true
+
+    private val sideloadedKitsCount: Int
+        get() {
+            try {
+                val kits = MParticle.getInstance()?.Internal()?.kitManager?.supportedKits
+                var count = 0
+                if (kits != null) {
```
einsteinx2, line 29: More of a style question, but is it more idiomatic to use `.let {}` instead of `if != null`? Are there any downsides to using `.let`? 

In Swift at least, it's considered non-idiomatic to do null checks this way anymore vs `if let variable = variable` which I think is similar in principle to the `.let{}` in kotlin.

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+            attributes.put(MessageKey.PRODUCT, Build.PRODUCT)
+            attributes.put(MessageKey.DEVICE, Build.DEVICE)
+            attributes.put(MessageKey.MANUFACTURER, Build.MANUFACTURER)
+            attributes.put(MessageKey.PLATFORM, operatingSystemString)
+            attributes.put(MessageKey.OS_VERSION, Build.VERSION.SDK)
+            attributes.put(MessageKey.OS_VERSION_INT, Build.VERSION.SDK_INT)
+            attributes.put(MessageKey.MODEL, Build.MODEL)
+            attributes.put(MessageKey.RELEASE_VERSION, Build.VERSION.RELEASE)
+
+            val application = appContext as Application
```
einsteinx2, line 171: Are we sure that `appContext` is always an `Application`? Can we add a check for that here and handle the case where it isn't? Or maybe better, if it's always an `Application` can we just change the input parameter to `(applicaiton: Application)` to ensure it?

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+            // locales
+            val locale = Locale.getDefault()
+            attributes.put(MessageKey.DEVICE_COUNTRY, locale.displayCountry)
+            attributes.put(MessageKey.DEVICE_LOCALE_COUNTRY, locale.country)
+            attributes.put(MessageKey.DEVICE_LOCALE_LANGUAGE, locale.language)
+            attributes.put(MessageKey.DEVICE_TIMEZONE_NAME, MPUtility.getTimeZone())
+            attributes.put(MessageKey.TIMEZONE, TimeZone.getDefault().rawOffset / (1000 * 60 * 60))
+            // network
+            val telephonyManager = appContext
+                .getSystemService(Context.TELEPHONY_SERVICE) as TelephonyManager
```
einsteinx2, line 199: Similar question here. What happens if this is not a `TelephonyManager` object? Is there something like `as?` in Kotlin or some other way to gracefully handle this if the cast fails?

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+            val phoneType = telephonyManager.phoneType
+            if (phoneType != TelephonyManager.PHONE_TYPE_NONE) {
```
einsteinx2, line 201: Nitpick, but seems like the extra variable assignment is unnecessary.

```suggestion
            if (telephonyManager.phoneType != TelephonyManager.PHONE_TYPE_NONE) {
```

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+        val adIdInfo = MPUtility.getAdIdInfo(context)
+        var message =
+            "Failed to collect Advertising ID, be sure to add Google Play services (com.google.android.gms:play-services-ads) or Amazon Ads (com.amazon.android:mobile-ads) to your app's dependencies."
+        if (adIdInfo != null) {
+            try {
+                deviceInfo.put(MessageKey.LIMIT_AD_TRACKING, adIdInfo.isLimitAdTrackingEnabled)
+                val instance = MParticle.getInstance()
+                // check instance nullability here and decline to act if it is not available. Don't want to have the case where we are overriding isLimiAdTrackingEnabled
+                // just because there was a timing issue with the singleton
+                if (instance != null) {
```
einsteinx2, line 248: Same question here about the null check vs the let syntax. Mainly trying to see if we can standardize on just try (assuming that's idiomatic).

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+            }
+        }
+        if (firstCollection) {
+            Logger.debug(message)
+            firstCollection = false
+        }
+
+        try {
+            val mParticle = MParticle.getInstance()
+            if (mParticle != null) {
```
einsteinx2, line 270: Same here

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
```diff
@@ -0,0 +1,343 @@
...
+        private var _deviceImei: String? = null
+
+        @get:JvmStatic
+        val deviceImei: String?
+            get() = _deviceImei ?: null
+
+        @JvmStatic
+        fun setDeviceImei(deviceImei: String?) {
+            _deviceImei = deviceImei
+        }
```
einsteinx2, line 331: Can this just be rewritten as a single property? What's the idea behind having a get only property with a private backing ivar and then a separate setter method?

einsteinx2 (changes requested):

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
Mansi-mParticle: We can use let{} here, but since we're using a for loop with the kits instance, we should use an if condition instead of let.

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
Mansi-mParticle: //check instance nullability here and decline to act if it is not available. Don't want to have the case where we are overriding isLimiAdTrackingEnabled
                //just because there was a timing issue with the singleton
                
The previous developer added this comment for this condition.

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
Mansi-mParticle: We can't use let here because, up until line 275, we'd have to check for null on every line if we did.

android-core/src/main/java/com/mparticle/internal/DeviceAttributes.kt
Mansi-mParticle: Since it’s a single property and needed to be changed in the existing method,So I created separate getter and setter for it.

einsteinx2 (approved):

PR #525, opened 2024-12-09
refactor: Migrate Internal MPUtility class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s MPUtility classe to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6718


android-core/src/main/java/com/mparticle/internal/MPUtility.kt
```diff
@@ -0,0 +1,906 @@
...
+        val resolveInfo: List<*> =
+            packageManager.queryIntentServices(
+                intent,
+                PackageManager.MATCH_DEFAULT_ONLY
+            )
+        return resolveInfo.size > 0
+    }
+
+    fun wrapExtras(extras: Bundle?): JSONObject? {
+        try {
```
einsteinx2, line 655: What was the purpose of adding the extra try catch around the whole method?

einsteinx2 (changes requested):

einsteinx2 (approved):

PR #527, opened 2024-12-10
refactor: Migrate Internal BaseHandler class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s BaseHandler classe to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6707


android-core/src/main/java/com/mparticle/internal/BaseHandler.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    fun removeMessage(what: Int) {
+        val messages: Set<Message?> = messageQueue.keys
+        for (message in messages) {
+            message?.let {
+                if (message.what == what) {
+                    messageQueue.remove(message)
+                }
+            }
+        }
```
einsteinx2, line 85:
```suggestion
        val messages: Set<Message> = messageQueue.keys
        for (message in messages) {
            if (message.what == what) {
                messageQueue.remove(message)
            }
        }
```

Looks like this doesn't need to be a set of optionals, so this can be simplified a bit

einsteinx2 (changes requested):

einsteinx2 (approved):

PR #530, opened 2024-12-17
ci: Update cross-platform-tests.yml

rmi22186:
## Instructions

 ## Summary
It looks like macos-12 has been deprecated, per the error here https://github.com/mParticle/mparticle-android-sdk/actions/runs/12261598641/job/34492232250?pr=526

image

 ## Testing Plan


Mansi-mParticle (approved):
Looks good to me

PR #533, opened 2024-12-20
refactor: Migrate Internal ProviderPersistence class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s ProviderPersistence classe to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6978


android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
```diff
@@ -0,0 +1,132 @@
...
+        private const val MACRO_TIMESTAMP = "%ts%"
+        private const val MACRO_GUID_LEAST_SIG = "%glsb%"
+
+        /**
+         * Macros are used so that the /config API call can come from a CDN (not user-specific).
+         */
+        private fun applyMacro(defaultString: String): String {
+            var defaultString = defaultString
+            if (!isEmpty(defaultString) && defaultString.startsWith("%")) {
+                defaultString = defaultString.lowercase(Locale.getDefault())
```
einsteinx2, line 95: Are we sure we want to use the default locale here? I'm not saying we don't, but is this different in any way from the `toLowerCase()` function in Java? Does `getDefault()` get the current system default or some universal default? I've been bitten by locale bugs in the past, so I want to make sure we're 100% certain here.

android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
```diff
@@ -0,0 +1,132 @@
...
+                } else if (defaultString == MACRO_GUID_LEAST_SIG) {
+                    return UUID.randomUUID().leastSignificantBits.toString()
+                }
+            }
+            return defaultString
+        }
+
+        private fun generateAID(): String {
+            var uuid = UUID.randomUUID().toString().replace("-", "")
+            uuid = uuid.uppercase(Locale.getDefault())
```
einsteinx2, line 113: Same concern about locale here.

einsteinx2 (changes requested):
Just those questions about locale, otherwise looks good to me.

android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
Mansi-mParticle: so here Locale.getDefault() will return the default language for system. languages where case rules vary (e.g., Turkish).

android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
```diff
@@ -0,0 +1,132 @@
...
+ * server side via mParticle. Rather than start from stratch, it's crucial that we can query data that
+ * the given SDK had been storing client-side.
+ */
+internal class ProviderPersistence(config: JSONObject, context: Context) : JSONObject() {
+    init {
+        val configPersistence = config.getJSONArray(KEY_PERSISTENCE)
+        for (i in 0 until configPersistence.length()) {
+            val values = JSONObject()
+            if (configPersistence.getJSONObject(i).has(KEY_PERSISTENCE_ANDROID)) {
+                val files = configPersistence.getJSONObject(i).getJSONArray(KEY_PERSISTENCE_ANDROID)
```
rmi22186, line 23: i suggest turn `configPersistence.getJSONObject(i)` into a variable for readability

android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
```diff
@@ -0,0 +1,132 @@
...
+                for (fileIndex in 0 until files.length()) {
+                    val fileObject = files.getJSONObject(fileIndex)
+                    val preferences =
+                        context.getSharedPreferences(fileObject.getString(KEY_PERSISTENCE_FILE), fileObject.getInt(KEY_PERSISTENCE_MODE))
+                    val fileObjects = fileObject.getJSONArray(KEY_PERSISTENCE_KEY_LIST)
+                    val editor = preferences.edit()
+                    for (keyIndex in 0 until fileObjects.length()) {
+                        val type = fileObjects.getJSONObject(keyIndex).getInt(KEY_PERSISTENCE_TYPE)
+                        val key = fileObjects.getJSONObject(keyIndex).getString(KEY_PERSISTENCE_KEY)
+                        val mpKey = fileObjects.getJSONObject(keyIndex).getString(KEY_PERSISTENCE_MPVAR)
```
rmi22186, line 34: i suggest turn `fileObjects.getJSONObject(keyIndex)` into a variable for readability

android-core/src/main/kotlin/com/mparticle/internal/ProviderPersistence.kt
```diff
@@ -0,0 +1,132 @@
...
+                } else if (defaultString == MACRO_GUID) {
+                    return UUID.randomUUID().toString()
+                } else if (defaultString == MACRO_TIMESTAMP) {
+                    return System.currentTimeMillis().toString()
+                } else if (defaultString == MACRO_GUID_LEAST_SIG) {
+                    return UUID.randomUUID().leastSignificantBits.toString()
+                }
+            }
+            return defaultString
+        }
```
rmi22186, line 109: there's a lot of else if statement's here, it looks like it can be updated to be more readable as follows, similar to what you did on line 42-46:

```
private fun applyMacro(defaultString: String): String {
    val lowerDefault = defaultString.lowercase(Locale.getDefault())
    return when {
        isEmpty(defaultString) || !lowerDefault.startsWith("%") -> defaultString
        lowerDefault == MACRO_GUID_NO_DASHES -> UUID.randomUUID().toString().replace("-", "")
        lowerDefault == MACRO_OMNITURE_AID -> generateAID()
        lowerDefault == MACRO_GUID -> UUID.randomUUID().toString()
        lowerDefault == MACRO_TIMESTAMP -> System.currentTimeMillis().toString()
        lowerDefault == MACRO_GUID_LEAST_SIG -> UUID.randomUUID().leastSignificantBits.toString()
        else -> defaultString
    }
}
```
(please check this work above if you agree)

rmi22186 (changes requested):

rmi22186 (approved):
LGTM.  can you make sure the tests are passing though?  might just need a rebase

PR #535, opened 2025-01-14
refactor: Migrate Internal Message class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s Message class to Kotlin.

 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6715

### DIFF
--- a/android-core/src/main/java/com/mparticle/internal/MessageHandler.kt
+++ b/android-core/src/main/java/com/mparticle/internal/MessageHandler.kt
@@ -46,9 +46,7 @@ internal open class MessageHandler : BaseHandler {
         mContext = context
         mMParticleDBManager = dbManager
         mDataplanId = dataplanId
-        if (dataplanVersion != null) {
-            mDataplanVersion = dataplanVersion
-        }
+        mDataplanVersion = dataplanVersion
     }
 
     constructor(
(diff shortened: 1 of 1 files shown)


android-core/src/main/java/com/mparticle/internal/MessageHandler.kt
```diff
@@ -0,0 +1,331 @@
...
+        dataplanId: String?,
+        dataplanVersion: Int?
+    ) {
+        mMessageManagerCallbacks = messageManager
+        mContext = context
+        mMParticleDBManager = dbManager
+        mDataplanId = dataplanId
+        if (dataplanVersion != null) {
+            mDataplanVersion = dataplanVersion
+        }
```
einsteinx2, line 51:
```suggestion
        mDataplanVersion = dataplanVersion
```

Seems like there's no need for a null check here since `mDataplanVersion` is nullable.

android-core/src/main/java/com/mparticle/internal/MessageHandler.kt
```diff
@@ -0,0 +1,331 @@
...
+    ) : super(looper) {
+        mMessageManagerCallbacks = messageManager
+        mContext = context
+        mMParticleDBManager = dbManager
+        if (dataplanId != null) {
+            mDataplanId = dataplanId
+        }
+        if (dataplanVersion != null) {
+            mDataplanVersion = dataplanVersion
+        }
```
einsteinx2, line 70:
```suggestion
        mDataplanId = dataplanId
        mDataplanVersion = dataplanVersion
```

Same here, these are both nullable so no null check needed.

einsteinx2 (changes requested):
Couple minor change suggestions, then after those are merged this can be approved.

einsteinx2 (approved):

PR #537, opened 2025-01-15
ci: update emulator version tag

Mansi-mParticle:
## Instructions

 ## Summary
 - Updated the GitHub Actions workflow to use the version tag 'v2' instead of a specific commit hash for emulator . 

 ## Testing Plan

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -82,7 +82,7 @@ jobs:
           java-version: "17"
           cache: "gradle"
       - name: "Run Instrumented Tests"
-        uses: reactivecircus/android-emulator-runner@842c587ad8aa4c68eeba24c396e15af4c2e9f30a #v2.33.0
+        uses: reactivecircus/android-emulator-runner@62dbb605bba737720e10b196cb4220d374026a6d #v2.33.0
         with:
           api-level: 29
           #script: ./gradlew :android-core:cAT :android-kit-base:cAT --stacktrace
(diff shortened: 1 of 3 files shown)


.github/workflows/daily.yml
```diff
@@ -82,16 +82,16 @@ jobs:
           java-version: "17"
           cache: "gradle"
       - name: "Run Instrumented Tests"
-        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b #v2.28.0
+        uses: reactivecircus/android-emulator-runner@842c587ad8aa4c68eeba24c396e15af4c2e9f30a #v2.33.0
```
rmi22186, line 85: this is different from the other commit hashes

.github/workflows/daily.yml
```diff
@@ -115,12 +115,12 @@ jobs:
           java-version: "17"
           cache: "gradle"
       - name: "Run Instrumented Orchestrator Tests"
-        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b #v2.28.0
+        uses: reactivecircus/android-emulator-runner@842c587ad8aa4c68eeba24c396e15af4c2e9f30a #v2.33.0
```
rmi22186, line 118: same comment.  should these both be 62dbb605bba737720e10b196cb4220d374026a6d?

rmi22186 (approved):

PR #538, opened 2025-01-24
refactor: Migrate Internal Config class to kotlin

Mansi-mParticle:
## Instructions

 ## Summary
 - Migrate internal package’s MPUtility classe to Kotlin.
 
 ## Testing Plan

 - Closes https://go.mparticle.com/work/SQDSDKS-6708


android-core/src/main/kotlin/com/mparticle/internal/ConfigManager.kt
```diff
@@ -0,0 +1,1436 @@
...
+            } else {
+                return null
+            }
+        }
+        set(timestamp) {
+            if (timestamp != null) {
+                sPreferences?.edit()
+                    ?.putLong(CONFIG_JSON_TIMESTAMP, timestamp)
+                    ?.apply()
+            }
```
einsteinx2, line 251: Isn't the logic here different from the original Java? In the original, `putLong` could be called with a null value.

The `putLong` method is defined as: `public abstract SharedPreferences.Editor putLong (String key, long value)` in Java, with `value` being the `long` value type. In the previous logic, if `timestamp` was null, would this have converted that to a `0` and stored it?

In the new Kotlin implementation, it will leave the previously stored value which may not be the desired logic.

einsteinx2: Note that I haven't tested this myself to confirm, so there may be no difference in the implementations in reality, but worth double checking.

android-core/src/main/kotlin/com/mparticle/internal/ConfigManager.kt
```diff
@@ -0,0 +1,1436 @@
...
+        }
+        set(log) {
+            sPreferences?.edit()?.putBoolean(Constants.PrefKeys.REPORT_UNCAUGHT_EXCEPTIONS, log)?.apply()
+        }
+    open val apiKey: String
+        get() = sPreferences?.getString(Constants.PrefKeys.API_KEY, null) ?: ""
+
+    open val apiSecret: String
+        get() = sPreferences?.getString(Constants.PrefKeys.API_SECRET, null) ?: ""
```
einsteinx2, line 528: I notice in the original Java we return null for both of these if it's not stored. Should these properties be nullable and return null instead of an empty string? Are there any null checks in other places that assume this can return null?

einsteinx2 (changes requested):
Couple of questions, otherwise looks good

android-core/src/main/kotlin/com/mparticle/internal/ConfigManager.kt
Mansi-mParticle: it is returning null it breaking lot's of existing code and test cases so decide to keep this as non null

BrandonStalnaker (dismissed):
LGTM but would appreciate another set of eyes to take a look

thomson-t (commented):
Do we want to merge the blackout-2024 branch first and then this one to main?

thomson-t:
Closing for now

PR #540, opened 2025-02-04
fix: Disable location tracking to stop setting location updates

Mansi-mParticle:
## Instructions

 ## Summary
 - Fix issue where previous location details were still being sent after calling `disableLocationTracking()`. Now, setLocation is set to null to clear the previous location.

 ## Testing Plan
 - Tested with sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-7031

### DIFF
--- a/android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
+++ b/android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
@@ -10,7 +10,6 @@ import com.mparticle.identity.IdentityApiRequest
 import com.mparticle.identity.IdentityStateListener
 import com.mparticle.internal.ConfigManager
 import com.mparticle.internal.KitFrameworkWrapper
-import com.mparticle.internal.Logger
 import com.mparticle.internal.MParticleJSInterface
 import com.mparticle.internal.MessageManager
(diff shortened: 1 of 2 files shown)


rmi22186 (commented):
can you add a test to this PR that shows the behavior after this change?

android-core/src/androidTest/kotlin/com.mparticle/MParticleTest.kt
```diff
@@ -329,6 +331,27 @@ class MParticleTest : BaseCleanStartedEachTest() {
...
+
+    @Test
+    fun testEnableLocationTracking() {
+        val location = Location("")
+        val mp = MParticle.getInstance()
+        mp!!.enableLocationTracking(LocationManager.NETWORK_PROVIDER, 30*1000, 1000)
+        mp!!.setLocation(location)
+        Assert.assertEquals(location, mp!!.mMessageManager.location)
+        Assert.assertNotNull(mp.mMessageManager.location)
+    }
```
rmi22186, line 353: This looks fine.  I would swap these tests, so that `testEnableLocationTracking` comes before `testEnableLocationTrackingAndDisableLocationTracking`. Not a huge deal, but I like to go from less specific to more specific tests.

rmi22186 (approved):

PR #550, opened 2025-03-26
feat: Add support for Rokt first-party kit

Mansi-mParticle:
## Instructions

 ## Summary
 - Added Rokt Inner class with a selectPlacements method that interacts with the Rokt SDK through the Rokt kit integrations.

 ## Testing Plan
 - Tested with sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-7154


android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1699,6 +1707,26 @@ public interface ResetListener {
     }
 
 
+    /**
+     * Rokt Integration
+     * */
+    public class Rokt{
```
sahil-rokt, line 1713: How about object using Rokt ?

android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1699,6 +1707,26 @@ public interface ResetListener {
...
+        protected Rokt(){
+
+        }
+        public void selectPlacements(String viewName,
+                            Map<String, String> attributes,
+                            Runnable onUnload,
+                            Runnable onLoad,
+                            Runnable onShouldHideLoadingIndicator,
+                            Runnable onShouldShowLoadingIndicator,
+                            Map<String, WeakReference<Object>> placeHolders,
```
sahil-rokt, line 1723: strong type is recommended, rather than using Object.

android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
```diff
@@ -646,6 +647,20 @@ public void reset() {
         }
     }
 
+    @Override
+    public void execute(String viewName,
+                        Map<String, String> attributes,
+                        Runnable onUnload,
+                        Runnable onLoad,
```
sahil-rokt, line 654: Want to confirm on which thread these Runnable executes because all the callbacks needs to execute on MainThread as partner updates their UI on the response.

android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1699,6 +1707,26 @@ public interface ResetListener {
...
 
+    /**
+     * Rokt Integration
+     * */
+    public class Rokt{
+        protected Rokt(){
+
+        }
+        public void selectPlacements(String viewName,
+                            Map<String, String> attributes,
```
rmi22186, line 1718: there should be a second method here also with just 2 arguments, per https://chat.google.com/room/AAAACu7-w_Q/9srmdeyLAiA/TmzM07mEqMA?cls=10 in order the simplify the invocation of this method by the partner

rmi22186: like brandon did here
https://github.com/mParticle/mparticle-apple-sdk/pull/339/files

android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1699,6 +1707,26 @@ public interface ResetListener {
...
+                            Runnable onUnload,
+                            Runnable onLoad,
+                            Runnable onShouldHideLoadingIndicator,
+                            Runnable onShouldShowLoadingIndicator,
+                            Map<String, WeakReference<Object>> placeHolders,
+                            Map<String, WeakReference<Typeface>> fontTypefaces) {
+             if (mConfigManager.isEnabled()) {
+                 mKitManager.execute(viewName,attributes, onUnload,onLoad,onShouldHideLoadingIndicator,onShouldShowLoadingIndicator,placeHolders,fontTypefaces);
+             }
```
rmi22186, line 1727: can we make this more readable by putting each argument on its own line?

android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
```diff
@@ -646,6 +647,20 @@ public void reset() {
...
+                        Runnable onShouldHideLoadingIndicator,
+                        Runnable onShouldShowLoadingIndicator,
+                        Map<String, WeakReference<Object>> placeHolders,
+                        Map<String, WeakReference<Typeface>> fontTypefaces) {
+        if (mKitManager != null) {
+            mKitManager.execute(viewName,attributes, onUnload, onLoad, onShouldHideLoadingIndicator, onShouldShowLoadingIndicator,placeHolders,fontTypefaces);
```
rmi22186, line 660: same comment about each parameter having its own line for readability

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1319,6 +1320,34 @@ public void reset() {
...
+                if (provider instanceof KitIntegration.RoktListener && !provider.isDisabled()) {
+                    MParticleUser user = MParticle.getInstance().Identity().getCurrentUser();
+                    Map<String, Object> objectAttributes = new HashMap<>();
+
+                    for (Map.Entry<String, String> entry : attributes.entrySet()) {
+                        objectAttributes.put(entry.getKey(), entry.getValue());
+                    }
+
+                    user.setUserAttributes(objectAttributes);
+                    ((KitIntegration.RoktListener) provider).execute(viewName,attributes, onUnload,onLoad,onShouldHideLoadingIndicator,onShouldShowLoadingIndicator,placeHolders,fontTypefaces,FilteredMParticleUser.getInstance(user.getId(), provider));
```
rmi22186, line 1343: same comment about each parameter getting its own line

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1319,6 +1320,34 @@ public void reset() {
...
+                }
+            } catch (Exception e) {
+                Logger.warning("Failed to call onLoginCompleted for kit: " + provider.getName() + ": " + e.getMessage());
```
rmi22186, line 1346:
```suggestion
                Logger.warning("Failed to call execute for kit: " + provider.getName() + ": " + e.getMessage());
```

rmi22186 (changes requested):

android-core/src/main/java/com/mparticle/MParticle.java
Mansi-mParticle: We're not using the Rokt dependency directly here, so we don't call the Rokt object directly

android-core/src/main/java/com/mparticle/MParticle.java
Mansi-mParticle: We're not using the Rokt dependency directly here, so we don't call the Rokt object directly so that's why here using object

sahil-rokt: Then how the Partner App will create the `Widget` from the Rokt-Kit dependency which uses Rokt package, and call the method from Mparticle workspace with object ?

android-core/src/main/java/com/mparticle/MParticle.java
sahil-rokt: I meant using object class as `object Rokt` because we are not creating object of this class

android-core/src/main/java/com/mparticle/Widget.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public Widget(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public Widget(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+    }
+
+    public void setDimensionCallback(RoktWidgetDimensionCallback callback) {
```
sahil-rokt, line 26: NIT: these callbacks are required for cross-platforms.

sahil-rokt (approved):

android-core/src/main/java/com/mparticle/Widget.java
```diff
@@ -0,0 +1,20 @@
+package com.mparticle;
```
rmi22186, line 1: Was on a call with SYD and they suggested we align this to be called `RoktEmbeddedView` instead of `Widget` so that we standardize the Rokt iOS/Android kits more.  Please confirm @sahil-rokt

sahil-rokt: Yes, we agreed to use the common in both platform called as `RoktEmbeddedView`

sahil-rokt (approved):

rmi22186 (approved):

PR #559, opened 2025-04-10
feat: Add sandbox mode to attributes

Mansi-mParticle:
## Instructions

 ## Summary
 - Added logic to include the sandbox mode flag in the attributes map. This flag is only added if it's not already present, with a default value of "false" when the environment value is null.

 ## Testing Plan
 - Tested With sample app and executed test cases

 - Closes https://go.mparticle.com/work/SQDSDKS-7206

### DIFF
--- a/android-core/src/main/kotlin/com/mparticle/internal/Constants.kt
+++ b/android-core/src/main/kotlin/com/mparticle/internal/Constants.kt
@@ -463,6 +463,9 @@ object Constants {
 
             //batch was mutated
             const val MODIFIED_BATCH: String = "mb"
+
+            //Sandbox mode for
+            const val SANDBOX_MODE_ROKT: String = "sandbox"
         }
     }
 
--- a/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
+++ b/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
@@ -29,6 +29,7 @@ import com.mparticle.consent.ConsentState;
 import com.mparticle.identity.IdentityApiRequest;
 import com.mparticle.identity.IdentityStateListener;
 import com.mparticle.identity.MParticleUser;
+import com.mparticle.internal.Constants;
 import com.mparticle.internal.CoreCallbacks;
 import com.mparticle.internal.KitManager;
 import com.mparticle.internal.KitsLoadedCallback;
@@ -51,6 +52,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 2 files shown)


rmi22186 (changes requested):
Please add tests for the following cases:
Can you add a test case for:

1. mock the environment and doesn't pass a sandbox attribute to ensure the environment variable is being used
2. mock the environment as development, but passes sandbox attribute as `false` and ensure `false` is set on `selectPlacements`

android-core/src/main/kotlin/com/mparticle/internal/Constants.kt
```diff
@@ -463,6 +463,9 @@ object Constants {
 
             //batch was mutated
             const val MODIFIED_BATCH: String = "mb"
+
+            //Sandbox mode for
```
rmi22186, line 467:
```suggestion
            //Sandbox mode for Rokt
```

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -843,8 +845,9 @@ class KitManagerImplTest {
             Pair("customerId", "55555"),
             Pair("country", "US")
         )
+
         manager.execute("Test", attributes, null, null, null)
-        Assert.assertEquals(5, attributes.size)
+        Assert.assertEquals(6, attributes.size)
```
rmi22186, line 850: This was updated to 6, but you don't add the sandbox attribute to look for here.

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -969,8 +974,9 @@ class KitManagerImplTest {
             Pair("minorcatid", "5-45555"),
             Pair("country", "US")
         )
+
         manager.execute("Test", attributes, null, null, null)
-        Assert.assertEquals(5, attributes.size)
+        Assert.assertEquals(6, attributes.size)
```
rmi22186, line 979: ditto

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -1031,13 +1037,200 @@ class KitManagerImplTest {
             Pair("customerId", "55555"),
             Pair("country", "US")
         )
+
+        manager.execute("Test", attributes, null, null, null)
+        Assert.assertEquals(6, attributes.size)
```
rmi22186, line 1042: ditto

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1359,18 +1366,27 @@ public void execute(String viewName,
...
+                            finalAttributes.put(mapTo, value);
                         }
                     }
                     Map<String, Object> objectAttributes = new HashMap<>();
-                    for (Map.Entry<String, String> entry : attributes.entrySet()) {
-                        objectAttributes.put(entry.getKey(), entry.getValue());
+                    for (Map.Entry<String, String> entry : finalAttributes.entrySet()) {
+                        if(!finalAttributes.containsKey(Constants.MessageKey.SANDBOX_MODE_ROKT)) {
+                            objectAttributes.put(entry.getKey(), entry.getValue());
+                        }
```
rmi22186, line 1378: if entry === sandbox_mode_rokt?

BrandonStalnaker (commented):
Logic looks good/matches iOS. Had one small question/suggestion on the code but good otherwise

rmi22186 (approved):

PR #570, opened 2025-04-16
feat: Add option to disable kits in MParticleOptions

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR introduces a new option in MParticleOptions that allows users to disable specific kits. When a kit is disabled, it will not be initialized, and no events will be sent from that kit.

 ## Testing Plan
 - Tested locally using a release build and executed the unit test cases.

 - Closes https://go.mparticle.com/work/SQDSDKS-7219

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -194,7 +194,7 @@ public class MParticleOptions {
     }
 
     /**
-     * Get list of filteredKits kits
+     * Get list of disabled kits
      *
      * @return
      * */
@@ -481,7 +481,7 @@ public class MParticleOptions {
         }
 
         /**
-         * Add Filter kits options
+         * Add disable kits option
          *
          * @param kits
          * @return
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -55,7 +55,7 @@ public class MParticleOptions {
     private DataplanOptions mDataplanOptions;
     private Map<Class, List<Configuration>> mConfigurations = new HashMap();
     private List<SideloadedKit> sideloadedKits = new ArrayList<>();
-    private List<Integer> filteredKits = new ArrayList<>();
+    private List<Integer> disabledKits = new ArrayList<>();
 
     private MParticleOptions() {
     }
@@ -148,7 +148,7 @@ public class MParticleOptions {
         this.mDataplanOptions = builder.dataplanOptions;
         this.mConfigurations = builder.configurations;
         this.sideloadedKits = builder.sideloadedKits;
-        this.filteredKits = builder.filteredKits;
+        this.disabledKits = builder.disabledKits;
     }
 
     /**
@@ -199,8 +199,8 @@ public class MParticleOptions {
      * @return
      * */
     @NonNull
-    public List<Integer> getFilteredKits() {
-        return filteredKits;
+    public List<Integer> getDisabledKits() {
+        return disabledKits;
     }
 
     /**
@@ -412,7 +412,7 @@ public class MParticleOptions {
         private Map<Class, List<Configuration>> configurations = new HashMap();
         private boolean isAppDebuggable;
         private List<SideloadedKit> sideloadedKits = new ArrayList<>();
-        private List<Integer> filteredKits = new ArrayList<>();
+        private List<Integer> disabledKits = new ArrayList<>();
 
         private Builder(Context context) {
             this.context = context;
@@ -487,8 +487,8 @@ public class MParticleOptions {
          * @return
          */
         @NonNull
-        public Builder filteredKits(@NonNull List<Integer> kits) {
-            filteredKits.addAll(kits);
+        public Builder disableKits(@NonNull List<Integer> kits) {
+            disabledKits.addAll(kits);
             return this;
         }
 
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -55,6 +55,7 @@ public class MParticleOptions {
     private DataplanOptions mDataplanOptions;
     private Map<Class, List<Configuration>> mConfigurations = new HashMap();
     private List<SideloadedKit> sideloadedKits = new ArrayList<>();
+    private List<Integer> filteredKits = new ArrayList<>();
 
     private MParticleOptions() {
     }
@@ -147,6 +148,7 @@ public class MParticleOptions {
         this.mDataplanOptions = builder.dataplanOptions;
         this.mConfigurations = builder.configurations;
         this.sideloadedKits = builder.sideloadedKits;
+        this.filteredKits = builder.filteredKits;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -55,6 +55,7 @@ public class MParticleOptions {
     private DataplanOptions mDataplanOptions;
     private Map<Class, List<Configuration>> mConfigurations = new HashMap();
     private List<SideloadedKit> sideloadedKits = new ArrayList<>();
+    private List<Integer> filteredKits = new ArrayList<>();
```
BrandonStalnaker, line 58: I'm sorry I didn't look at the changes they suggested for android. They named this disabledKits in iOS and filteredKits for Android? We should align on one and I'm biased but I think it should be "disabledKits"

android-kit-base/src/main/java/com/mparticle/kits/KitIntegrationFactory.java
```diff
@@ -109,6 +109,7 @@ private void loadSideloadedIntegrations(MParticleOptions options) {
     }
 
     private void loadIntegrations(MParticleOptions options) {
+        filterKits(options);
```
BrandonStalnaker, line 112: Will this also prevent the kit from getting started or does it just filter kits out from having events sent to them. The iOS code they provided did both.

BrandonStalnaker (commented):
Logic seems fine and the tests are great just have some questions so we can be sure to be aligned across iOS and Android

android-core/src/main/java/com/mparticle/MParticleOptions.java
rmi22186: I agree with Brandon here.  Let's use the term `disable` throughout.  It is much more clear in its purpose than filter, which is a bit vague.

android-kit-base/src/main/java/com/mparticle/kits/KitIntegrationFactory.java
rmi22186: Based on what I'm seeing, `knownIntegrations` starts with all possible kits, this removes them from the list, and only any kit that is included as a dependency and in this list will get initialized.  therefor by not initializing a kit, it wouldn't have the option of having an event log to it either, do I have that right @Mansi-mParticle ?

Mansi-mParticle: Yes, that's right @rmi22186  — the kit will be removed before it's prepared for initialization

android-kit-base/src/main/java/com/mparticle/kits/KitIntegrationFactory.java
```diff
@@ -119,6 +120,13 @@ private void loadIntegrations(MParticleOptions options) {
         }
     }
 
+    private void filterKits(MParticleOptions options) {
+        for (Integer filteredKit : options.getDisabledKits()) {
```
alexs-mparticle, line 124: I think you should have a check to see if `options.getDisabledKits()` is empty and then just exit early.

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -602,6 +602,127 @@ class KitManagerImplTest {
...
     }
 
+    @Test
+    fun shouldFilterKitsFromKnownIntegrations() {
+        val options = MParticleOptions.builder(MockContext()).build()
+        val filteredKitOptions = MParticleOptions.builder(MockContext())
+            .disableKits(listOf(MParticle.ServiceProviders.ADJUST))
+            .disableKits(listOf(MParticle.ServiceProviders.APPBOY))
+            .disableKits(listOf(MParticle.ServiceProviders.CLEVERTAP))
+            .build()
```
rmi22186, line 612: Is this API consistent with the Android SDK options?  I was expecting instead of 3 lines of 
```
.disableKits(listOf(MParticle.ServiceProviders.ADJUST))
.disableKits(listOf(MParticle.ServiceProviders.APPBOY))
.disableKits(listOf(MParticle.ServiceProviders.CLEVERTAP))
```

to instead be more like iOS where you would just drop a single list like:

something like
```
.setFilteredKits(Arrays.asList(
        MParticle.ServiceProviders.ADJUST,
        MParticle.ServiceProviders.APPBOY,
        MParticle.ServiceProviders.CLEVERTAP
    ))
    .build();
```

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -602,6 +602,127 @@ class KitManagerImplTest {
...
+        Assert.assertNotNull(withoutFilterIntegration[MParticle.ServiceProviders.CLEVERTAP])
+
+        // Filtered kits; the specified kit should not be present (should be null)
+        Assert.assertNull(filteredKitIntegrations[MParticle.ServiceProviders.ADJUST])
+        Assert.assertNull(filteredKitIntegrations[MParticle.ServiceProviders.APPBOY])
+        Assert.assertNull(filteredKitIntegrations[MParticle.ServiceProviders.CLEVERTAP])
+    }
+
+    @Test
+    fun shouldFilterKitsFromKnownIntegrations_When_filter_Is_Empty() {
```
rmi22186, line 637: any reason to go from camelCase to snake_case here?

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -602,6 +602,127 @@ class KitManagerImplTest {
...
+        val knownKitsSize = knownIntegrations.size
+        val filteredKnownKitsSize = filteredKnownIntegrations.size
+        Assert.assertEquals(knownKitsSize, filteredKnownKitsSize)
+    }
+
+    @Test
+    fun shouldIgnoreUnknownKitInFilter() {
+        val options = MParticleOptions.builder(MockContext()).build()
+        val filteredKitOptions = MParticleOptions.builder(MockContext())
+            .disableKits(listOf(1231, 132132))
```
rmi22186, line 660: i see this API is used here, which makes more sense to me than the one above, so it seems like both APIs work for 
```
        val filteredKitOptions = MParticleOptions.builder(MockContext())
            .filteredKits(listOf(MParticle.ServiceProviders.ADJUST))
            .filteredKits(listOf(MParticle.ServiceProviders.APPBOY))
            .filteredKits(listOf(MParticle.ServiceProviders.CLEVERTAP))
```

and

```
.disableKits(listOf(
     MParticle.ServiceProviders.ADJUST, 
     MParticle.ServiceProviders.APPBOY,
     MParticle.ServiceProviders.CLEVERTAP
     )
)
```

is that correct?

android-kit-base/src/main/java/com/mparticle/kits/KitIntegrationFactory.java
Mansi-mParticle: Since getDisabledKits() never returns null and just gives an empty list, the loop won't run if there are no kits. Do you thing still need to check here?

rmi22186 (approved):

PR #574, opened 2025-04-24
feat: Add user attributes mapping logic in Rokt execute flow

Mansi-mParticle:
## Instructions

 ## Summary
 - Implement user attribute transformation based on placement mapping in Rokt execute

 ## Testing Plan
 - Tested With sample app and Executed unit test cases

 - Closes https://go.mparticle.com/work/SQDSDKS-7185

### DIFF
--- a/android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
+++ b/android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
@@ -61,6 +61,7 @@ public class KitConfiguration {
     private final static String KEY_CONSENT_FORWARDING_RULES_VALUE_CONSENTED = "c";
     private final static String KEY_CONSENT_FORWARDING_RULES_VALUE_HASH = "h";
     private final static String KEY_EXCLUDE_ANONYMOUS_USERS = "eau";
+    private final static String KEY_PLACEMENT_ATTRIBUTES = "placementAttributes";
 
     //If set to true, our sdk honor user's optout wish. If false, we still collect data on opt-ed out users, but only for reporting.
     private final static String HONOR_OPT_OUT = "honorOptOut";
@@ -984,6 +985,14 @@ public class KitConfiguration {
             return Boolean.parseBoolean(optOut);
         }
         return true;
+    }
+
+    public JSONArray getPlacementAttributes() throws JSONException {
+        if (!settings.containsKey(KEY_PLACEMENT_ATTRIBUTES)) {
+            return new JSONArray();
+        }
+        String jsonArrayStr = settings.get(KEY_PLACEMENT_ATTRIBUTES);
+        return new JSONArray(jsonArrayStr);
     }
 
     boolean shouldSetIdentity(MParticle.IdentityType identityType) {
--- a/android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
+++ b/android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
@@ -61,7 +61,7 @@ public class KitConfiguration {
     private final static String KEY_CONSENT_FORWARDING_RULES_VALUE_CONSENTED = "c";
     private final static String KEY_CONSENT_FORWARDING_RULES_VALUE_HASH = "h";
     private final static String KEY_EXCLUDE_ANONYMOUS_USERS = "eau";
-    private final static String KEY_PLACEMENT_ATTRIBUTES = "placementAttributes";
+    private final static String KEY_PLACEMENT_ATTRIBUTES_MAPPING = "placementAttributesMapping";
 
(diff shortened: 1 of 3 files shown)


android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -986,6 +987,14 @@ public boolean shouldHonorOptOut() {
         return true;
     }
 
+    public JSONArray getPlacementAttributes() throws JSONException {
+        if (!settings.containsKey(KEY_PLACEMENT_ATTRIBUTES)) {
```
BrandonStalnaker, line 991: Settings is the whole config from the server right? Not specifically the Rokt KitConfig where KEY_PLACEMENT_ATTRIBUTES is a key. Does this work?

BrandonStalnaker: Here's where I'm doing this in iOS. I'm iterating over all the kitConfigs we have to find the Rokt kit and use its config https://github.com/mParticle/mparticle-apple-sdk/pull/343/files#diff-16f0b9f48e63f6d2da9781559af2c9ce3a10add2960849aca319a5854aea2d3dR267

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1334,12 +1334,25 @@ public void execute(String viewName,
...
+
+                    for (int i = 0; i < jsonArray.length(); i++) {
+                        JSONObject obj = jsonArray.optJSONObject(i);
+                        if (obj == null) continue;
+
+                        String mapKey = obj.optString("map");
+                        String valueValue = obj.optString("value");
+                        if (attributes.containsKey(mapKey)) {
+                            String value = attributes.remove(mapKey);
+                            attributes.put(valueValue, value);
```
BrandonStalnaker, line 1348: maybe this might be more readable?
`mapkey` to `mapFrom`
`valueValue` to `mapTo`

BrandonStalnaker (changes requested):
My main question is around how you're getting the placements mapping

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
alexs-mparticle: I think I'm doing something similar on Web: https://github.com/mParticle/mparticle-web-sdk/blob/master/src/mp-instance.ts#L1398

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1334,12 +1334,25 @@ public void execute(String viewName,
             try {
                 if (provider instanceof KitIntegration.RoktListener && !provider.isDisabled()) {
                     MParticleUser user = MParticle.getInstance().Identity().getCurrentUser();
-                    Map<String, Object> objectAttributes = new HashMap<>();
 
+                    JSONArray jsonArray = provider.getConfiguration().getPlacementAttributes();
```
alexs-mparticle, line 1338: It seems to me that there's a lot of small steps going on in this function, but if something fails at any point during the execution, we'll just get a generic `Failed to call execute for kit` error message, which to me seems like it would be time consuming to investigate. Would it make sense to have some validation steps in between each operation and then throw a specific error message?

Such as if the `jsonArray` is empty, or if the `JSONObject obj = jsonArray.optJSONObject(i);` throws, we can have a specific `JSON parsing error`. Or if `attributes` is empty or invalid, we can have a specific message for that?

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1334,12 +1334,25 @@ public void execute(String viewName,
...
+
+                    for (int i = 0; i < jsonArray.length(); i++) {
+                        JSONObject obj = jsonArray.optJSONObject(i);
+                        if (obj == null) continue;
+
+                        String mapKey = obj.optString("map");
+                        String valueValue = obj.optString("value");
```
alexs-mparticle, line 1345: `valueValue` is rather confusing. Maybe you can clarify the shape of the data in a comment so it's more clear?

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
Mansi-mParticle: Already discuss over the call

BrandonStalnaker (approved):
Thanks for answer

alexs-mparticle (approved):

BrandonStalnaker (approved):

PR #575, opened 2025-04-29
feat: Update Android Gradle plugin to 8.1.0

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR updates the Android Gradle Plugin to version 8.1.0.

 ## Testing Plan
 - Tested it with a sample app and ran the test cases.

 - Closes https://go.mparticle.com/work/SQDSDKS-7068

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,10 @@ subprojects {
         if (project.hasProperty('android')) {
             project.android {
                 if (namespace == null) {
-                    namespace project.group
+                    throw new GradleException(
+                        "Module '${project.name}' is missing an explicit namespace declaration. " +
+                        "Please add 'namespace' to the android {} block in ${project.buildFile.name}"
+                    )
                 }
             }
         }
--- a/kit-plugin/src/main/groovy/com/mparticle/kits/KitPlugin.groovy
+++ b/kit-plugin/src/main/groovy/com/mparticle/kits/KitPlugin.groovy
@@ -140,11 +140,18 @@ class KitPlugin implements Plugin<Project> {
             }
 
             target.extensions.configure(SigningExtension) { signing ->
-                signing.required = { target.gradle.taskGraph.hasTask('publishReleasePublicationToMavenRepository') }
+                signing.required {
+                    target.gradle.taskGraph.hasTask("publishReleasePublicationToMavenRepository")
+                }
+
                 def signingKey = target.findProperty('signingKey') ?: System.getenv('mavenSigningKeyId')
                 def signingPassword = target.findProperty('signingPassword') ?: System.getenv('mavenSigningKeyPassword')
+
                 if (signingKey && signingPassword) {
-                    signing.useInMemoryPgpKeys(target.property('signingKey'), target.property('signingPassword'))
+                    signing.useInMemoryPgpKeys(signingKey, signingPassword)
+                    signing.sign(target.extensions.getByType(PublishingExtension)
+                        .publications
+                        .findByName("release"))
                 }
             }
         }
--- a/scripts/maven.gradle
+++ b/scripts/maven.gradle
@@ -124,5 +124,5 @@ afterEvaluate {
 }
 
 //Publishing task aliases for simpler local development
-task publishLocal { dependsOn "publishDebugPublicationToMavenLocal"}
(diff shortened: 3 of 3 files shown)


BrandonStalnaker (dismissed):
LGTM

android-core/build.gradle
```diff
@@ -13,8 +13,8 @@ android {
         exclude 'META-INF/LICENSE'
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
```
samdozor, line 16: how does this impact clients? do they need to build their app with 17 when compiling against our SDK? is that ok?

gradle.properties
```diff
@@ -2,4 +2,7 @@
 android.useAndroidX=true
 org.gradle.daemon=true
 org.gradle.jvmargs=-Xmx2560m
-#-XX:ThreadStackSize=4096 -XX:CompilerThreadStackSize=4096
\ No newline at end of file
+#-XX:ThreadStackSize=4096 -XX:CompilerThreadStackSize=4096
+android.defaults.buildfeatures.buildconfig=true
+org.gradle.caching=true
```
samdozor, line 7: is this going to mess up CI builds? we shouldn't be caching for CI

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Oct 04 12:58:48 PDT 2021
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-all.zip
```
samdozor, line 3: did you intentionally change it to all rather than just the binary? why?

samdozor (changes requested):

gradle.properties
Mansi-mParticle: I removed org.gradle.caching=true because the build cache can affect CI

android-core/build.gradle
Mansi-mParticle: Yes, this can affect the client app, but Android Gradle Plugin 8 requires at least Java 17 to run.

build.gradle
```diff
@@ -1,6 +1,6 @@
 buildscript {
     ext.kotlin_version = '1.9.0'
-    ext.gradle_version = '7.3.1'
+    ext.gradle_version = '8.1.0'
```
thomson-t, line 3: Can we use `8.1.4` so that it has the patch fixes as well? https://developer.android.com/build/releases/past-releases/agp-8-1-0-release-notes#android-gradle-plugin-8.1.4

kit-plugin/src/main/groovy/com/mparticle/kits/KitPlugin.groovy
```diff
@@ -139,17 +139,13 @@ class KitPlugin implements Plugin<Project> {
...
-            def signingPassword = System.getenv("mavenSigningKeyPassword")
-
-            if (signingKey != null) {
-                target.extensions.add('signing.keyId', signingKey)
-                target.extensions.add('signing.password', signingPassword)
-
-                SigningExtension signing = new SigningExtension(target)
-                signing.required = { target.gradle.taskGraph.hasTask("publishReleasePublicationToMavenRepository") }
-                signing.useInMemoryPgpKeys(signingKey, signingPassword)
-                signing.sign publishing.publications.findByName("release")
```
thomson-t: Not sure if the change is doing the same. Is the intention to sign it only when the key is provided and sign just the release variant?

build.gradle
```diff
@@ -26,6 +26,15 @@ sonarqube {
 }
 
 subprojects {
+    afterEvaluate { project ->
+        if (project.hasProperty('android')) {
+            project.android {
+                if (namespace == null) {
+                    namespace project.group
```
thomson-t: Is this safe to add this way?
Ideally we should be adding a namespace to each module's gradle file, right?

PR #578, opened 2025-05-20
feat: Fire Identify if Provided Email in Rokt SelectPlacements

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR adds logic to check the email passed to MParticle.Rokt.selectPlacements. Check if that email matches the one in the current mParticle user identity. If the email is missing or different, use the Identity modify API to update the email in the user's identity.

 ## Testing Plan
 - Tested with sample app and executed test cases.

 - Closes https://go.mparticle.com/work/SQDSDKS-7248

### DIFF
--- a/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
+++ b/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
@@ -1,5 +1,7 @@
 package com.mparticle.kits;
 
+import static android.util.Log.e;
+
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
@@ -10,6 +12,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.Looper;
+import android.util.Log;
 
 import androidx.annotation.MainThread;
 import androidx.annotation.NonNull;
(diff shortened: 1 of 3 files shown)


android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1361,13 +1375,61 @@ public void execute(String viewName,
...
+                                    "Please make sure to sync the email identity to mParticle as soon as it's available. " +
+                                    "Identifying user with the provided email before continuing to selectPlacements.",
+                            existingEmail, email
+                    ));
+                }
+
+                IdentityApiRequest identityRequest = IdentityApiRequest.withUser(user)
+                        .email(email)
+                        .build();
+                MParticleTask<IdentityApiResult> task = instance.Identity().identify(identityRequest);
```
alexs-mparticle, line 1409: I see you're passing in the entire `instance` to just use Identity. Would it be cleaner to just pass in the `identity` service by itself?

I know we do this a lot in our SDKs, where we pass in the entire mPInstance object, but we talked about this being an antipattern that we wanted to avoid.

alexs-mparticle (approved):
LGTM

BrandonStalnaker (approved):
LGTM

PR #582, opened 2025-06-11
feat: Passthrough SDK type to Rokt Kit

jamesnrokt:
## Instructions

 ## Summary
- Passthrough wrapper SDK type to Rokt kit

 ## Testing Plan

 - Closes https://go.mparticle.com/work/sqdsdks-7391


rmi22186 (approved):
@jamesnrokt - similar situation to iOS, Mansi removed the setWrapper call from MPRokt. since we can do it at the higher mParticle layer.  we are only calling it internally ever, so it doesn't need to be on MPRokt explicitly.  Having it here would require someone to call it on both MP and MPRokt in order for both SDKs to understand it.  Yes the MP layer isn't doing anything with it, but in the future we want it to.

PR #583, opened 2025-06-11
feat: Add support purchaseFinalized API in android core

Mansi-mParticle:
## Instructions

 ## Summary
 - Add support for Rokt's purchaseFinalized API in the Android core.

 ## Testing Plan
 - Tested with sample app and Added test cases.
 
 - Closes https://go.mparticle.com/work/SQDSDKS-7373

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticle.java
+++ b/android-core/src/main/java/com/mparticle/MParticle.java
@@ -1877,7 +1877,6 @@ public class MParticle {
 
         }
 
-
         public void selectPlacements(@NonNull String viewName,
                                      @NonNull Map<String, String> attributes,
                                      @Nullable MpRoktEventCallback callbacks,
@@ -1893,6 +1892,7 @@ public class MParticle {
                          config);
              }
         }
+
         public void selectPlacements(@NonNull String viewName,
                                      @NonNull Map<String, String> attributes) {
             if (mConfigManager.isEnabled()) {


android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1904,6 +1905,12 @@ public void selectPlacements(@NonNull String viewName,
                 );
             }
         }
+
+        public void reportConversion(@NonNull String placementId, @NonNull String catalogItemId, boolean status) {
```
rmi22186, line 1909: revert to `purchaseFinalized`

android-core/src/main/java/com/mparticle/MParticle.java
```diff
@@ -1877,6 +1877,7 @@ protected Rokt(){
 
         }
 
+
```
rmi22186, line 1880:
```suggestion
```

android-core/src/test/kotlin/com/mparticle/MParticleTest.kt
```diff
@@ -531,6 +531,31 @@ class MParticleTest {
...
+    @Test
+    fun testReportConversion_withBasicParams_whenEnabled() {
+        var instance: MParticle = InnerMockMParticle()
+        MParticle.setInstance(instance)
+        `when`(instance.mConfigManager.isEnabled()).thenReturn(true)
+
+        val attributes = mutableMapOf<String, String>()
+        attributes.put("a", "b")
+
+        instance.rokt.reportConversion("132", "1111", true)
```
rmi22186, line 543: revert to `purchaseFinalized`

android-core/src/test/kotlin/com/mparticle/MParticleTest.kt
```diff
@@ -531,6 +531,31 @@ class MParticleTest {
...
+        verify(instance.mKitManager).purchaseFinalized("132", "1111", true)
+    }
+
+    @Test
+    fun testReportConversion_withBasicParams_whenDisabled() {
+        var instance: MParticle = InnerMockMParticle()
+        MParticle.setInstance(instance)
+        `when`(instance.mConfigManager.isEnabled()).thenReturn(false)
+
+        instance.rokt.reportConversion("132", "1111", true)
```
rmi22186, line 554: revert to `purchaseFinalized`

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -1451,6 +1451,10 @@ class KitManagerImplTest {
         ) {
             println("Executed with $attributes")
         }
+
+        override fun purchaseFinalized(placementId: String, catalogItemId: String, status: Boolean) {
+            TODO("Not yet implemented")
```
rmi22186, line 1456: This is a nonblocking comment, but rather for my knowledge.

Can you explain this construct to me of having TODO("Not yet implemented") in various methods in this file?  I see that is't not just this one, so you are being consistent, but I also see `println('Executed with $attributes)` [right above here](https://github.com/mParticle/mparticle-android-sdk/pull/583/files#diff-f0dbae0f2ddfdc859d63ae13e8e367d47f460ba2e533623da17855c83d3bf067R1443-R1453).

I assume it's used to not break the test?  Why do `println` in one place, and `TODO` in another/

rmi22186 (approved):

thomson-t (approved):

PR #585, opened 2025-06-16
fix: ensure user attributes are applied before proceeding with Rokt execution

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR update makes sure that user attributes are fully saved before moving to the next step.
    It uses a callback with getUserAttributes to wait until the data is ready.
This helps avoid problems where Rokt tries to use user info before it's updated.

 ## Testing Plan
 - Tested with sample app


rmi22186 (changes requested):
minor nit

PR #588, opened 2025-06-18
fix: Include missing overridesSubdirectory flag in networkOptions for domain mapping

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR fixes an issue related to domain mapping.
When we implemented workspace switching, we added an "uploaded settings" option under networkOptions. However, the overridesSubdirectory field was missing from the networkOptions JSON, so it always defaulted to false.
As a result, when domain mapping required overridesSubdirectory to be true, it caused issues. This PR adds that missing field to prevent those problems.

 ## Testing Plan
 - Tested with sample app 

 - Closes https://go.mparticle.com/work/SQDSDKS-7377


BrandonStalnaker (approved):
LGTM Thanks for the explanation in the PR description!

PR #589, opened 2025-06-19
feat: Add support for the fontFilePathMap

Mansi-mParticle:
## Instructions

 ## Summary
 - This PR introduces support for passing font file information to the Rokt Kit via the Rokt.init method.
A new roktOptions field is added to the MParticle options, allowing the client to provide either the font's PostScript name or the file path. These options are forwarded to the Rokt Kit and used during its initialization.

 ## Testing Plan
 - Tested this feature with sample app.

 - Closes https://go.mparticle.com/work/SQDSDKS-7413


android-core/src/main/java/com/mparticle/internal/KitFrameworkWrapper.java
```diff
@@ -514,7 +515,15 @@ public void updateDataplan(@Nullable MParticleOptions.DataplanOptions dataplanOp
...
+    @Override
     @NonNull
+    public RoktOptions getRoktOptions() {
+        if (mKitManager != null) {
+            return mKitManager.getRoktOptions();
+        }
+        return mOptions != null ? mOptions.getRoktOptions() : new RoktOptions();
+    }
+
     @Override
```
rmi22186, line 527: Looks like the @NonNull is missing from `getKitStatus` now

rmi22186 (changes requested):

rmi22186 (approved):

PR #590, opened 2025-06-24
fix: Restore ProGuard rules to preserve Rokt class members

thomson-t:
## Instructions

 ## Summary
- `events` API from Rokt class was getting removed by proguard. Keep the class members and annotations for kotlin metadata

 ## Testing Plan
 - Verified the local release snapshot aar file 
Screenshot 2025-06-24 at 10 35 49 am

### DIFF
--- a/android-core/proguard.pro
+++ b/android-core/proguard.pro
@@ -208,11 +208,21 @@
 -keep class com.mparticle.rokt.RoktLayoutDimensionCallBack { *; }
 -keep class com.mparticle.rokt.RoktOptions { *; }
 
+# Preserve annotation classes
+-keep class androidx.annotation.NonNull { *; }
+-keep class androidx.annotation.Nullable { *; }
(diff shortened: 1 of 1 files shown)


android-core/proguard.pro
```diff
@@ -208,10 +208,17 @@
...
+    @androidx.annotation.NonNull *;
+    @androidx.annotation.Nullable *;
+}
+
 # Preserve all method signatures in the Rokt class to prevent overload resolution issues
 -keepclassmembers class com.mparticle.Rokt {
-    public void selectPlacements(...);
-    public void purchaseFinalized(...);
+    *;
 }
```
jamesnrokt, line 222:
```suggestion
    *;
}
-keepclassmembers class com.mparticle.KitIntegration$RoktListener* {
     *;
 }
```

jamesnrokt (approved):

Mansi-mParticle (approved):
LGTM

PR #592, opened 2025-07-01
fix: Refactor InstallReferrerClient to use reflection

jamesnrokt:
## Instructions

 ## Summary
 - Fixes a fatal crash on applications built with frameworks like .NET MAUI. The crash is a NoClassDefFoundError that occurs when the InstallReferrerHelper class is accessed.
 
```
NoClassDefFoundError: Failed resolution of: Lcom/android/installreferrer/api/InstallReferrerClient;
    at com.mparticle.InstallReferrerHelper$1.run(InstallReferrerHelper.java:49)
```

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why.
 - Used sample app to ensure that install receiver was provided. Received `utm_source=google-play&utm_medium=organic`
 - 

 - Closes https://go.mparticle.com/work/SQDSDKS-7474


android-core/src/main/java/com/mparticle/InstallReferrerHelper.java
```diff
@@ -3,34 +3,41 @@
...
 import android.content.SharedPreferences;
 import android.os.Looper;
-import android.os.RemoteException;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
-import com.android.installreferrer.api.InstallReferrerClient;
-import com.android.installreferrer.api.InstallReferrerStateListener;
-import com.android.installreferrer.api.ReferrerDetails;
```
thomson-t, line 13: There is a `compileOnly` dependency defined [here](https://github.com/mParticle/mparticle-android-sdk/blob/main/android-core/build.gradle#L142) for `installreferrer`. If we are moving this to reflection methods, we could remove that as well.

thomson-t (dismissed):

thomson-t (approved):

Mansi-mParticle (approved):
LGTM

PR #598, opened 2025-07-24
fix: Print domain mapping logs only once per session

Mansi-mParticle:
## Instructions

 ## Summary
 - Before, NetworkConfigure was called only once in a session. But after adding workspace switching, it started calling NetworkConfigure for every upload. Because of that, the domain mapping logs were printed many times in the same session.
To fix this, I added a HashSet to make sure the domain mapping logs print only once per session.

 ## Testing Plan
 - Tested with sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-7554

### DIFF
--- a/android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
+++ b/android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
@@ -20,12 +20,15 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.HashSet;
+import java.util.Set;
 
 public class NetworkOptions {
 
     Map<Endpoint, DomainMapping> domainMappings = new HashMap<Endpoint, DomainMapping>();
     boolean pinningDisabledInDevelopment = false;
     boolean pinningDisabled = false;
+    private static Set<Endpoint> loggedDomainTypes = new HashSet<>();
 
     private NetworkOptions() {
     }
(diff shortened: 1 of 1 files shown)


android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
```diff
@@ -144,10 +147,13 @@ public Builder addDomainMapping(@Nullable DomainMapping domain) {
...
-                try {
-                    Logger.warning("Duplicate DomainMapping submitted, DomainMapping:\n" + domain.toJson().toString(4) + "\n will overwrite DomainMapping:\n" + domain.toJson().toString(4));
-                } catch (JSONException e) {
-                    e.printStackTrace();
+                if (!loggedDomainTypes.contains(domain.getType())) {
+                    try {
+                        Logger.warning("Duplicate DomainMapping submitted, DomainMapping:\n" + domain.toJson().toString(4) + "\n will overwrite DomainMapping:\n" + domain.toJson().toString(4));
+                        loggedDomainTypes.add(domain.getType());
+                    } catch (JSONException e) {
+                        e.printStackTrace();
```
thomson-t, line 155: Not sure if this is ideal, shouldn't this be 
```suggestion
                        Logger.error(e);
```

thomson-t (approved):

rmi22186 (approved):
lgtm pending confirming thomsons comment

PR #599, opened 2025-07-28
feat: Handle hashed email for rokt

Mansi-mParticle:
## Instructions

 ## Summary
 -  This PR includes changes to handle `other` before calling the Identity API and invoking the Rokt Kit.

 ## Testing Plan
 - Tested with sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-7556


rmi22186 (approved):
logic looks good.  Not sure why there were a lot of mockito updates to the rest of the file, but LGTM overall

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1416,6 +1417,15 @@ public void execute(@NonNull String viewName,
         }
     }
 
+    public static String getValueIgnoreCase(Map<String, String> map, String searchKey) {
```
rmi22186, line 1420: Can you add tests for `getValueIgnoreCase`?

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1416,6 +1417,15 @@ public void execute(@NonNull String viewName,
...
+        for (Map.Entry<String, String> entry : map.entrySet()) {
+            if (entry.getKey().equalsIgnoreCase(searchKey)) {
+                return entry.getValue();
```
rmi22186, line 1423: Should this return the `searchKey` instead?

Because if we are searching for `email`, and the entry value is actually `EMAIL`, then `EMAIL` would be returned, right?

BrandonStalnaker (approved):
LGTM

rmi22186 (approved):

PR #601, opened 2025-08-19
feat: map selected identity 'emailsha256'

Mansi-mParticle:
## Instructions

 ## Summary
 - Updated the logic to use the selected identity as the emailsha256 for Rokt. If the selected identity isn’t available, it will default to 'other' until the backend changes are live.

 ## Testing Plan
 - Tested with sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-7629

### DIFF
--- a/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
+++ b/android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
@@ -1528,11 +1528,11 @@ public class KitManagerImpl implements KitManager, AttributionListener, UserAttr
                 String identityTypeStr = (kitConfiguration != null)
                         ? kitConfiguration.getHashedEmailUserIdentityType()
                         : null;
-                if (identityTypeStr != null && !identityTypeStr.equalsIgnoreCase("Unknown")) {
+                if (identityTypeStr != null ) {
                     selectedIdentityType = MParticle.IdentityType.valueOf(identityTypeStr);
                 }
             } catch (IllegalArgumentException e) {
-                selectedIdentityType = MParticle.IdentityType.Other; // fallback if invalid value
+                Logger.error("Invalid identity type "+e.getMessage());
             }
             String existingEmail = user.getUserIdentities().get(MParticle.IdentityType.Email);
             String existingHashedEmail = selectedIdentityType != null ? user.getUserIdentities().get(selectedIdentityType) : null;
@@ -1563,7 +1563,7 @@ public class KitManagerImpl implements KitManager, AttributionListener, UserAttr
                 if (emailMismatch) {
                     identityBuilder.email(email);
                 }
-                if (hashedEmailMismatch && selectedIdentityType != null) {
+                if (hashedEmailMismatch) {
                     identityBuilder.userIdentity(selectedIdentityType, hashedEmail);
                 }
 


android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1550,8 +1553,14 @@ else if (hashedEmailMismatch && existingHashedEmail != null) {
                 if (emailMismatch) {
                     identityBuilder.email(email);
                 }
-                if (hashedEmailMismatch) {
-                    identityBuilder.userIdentity(MParticle.IdentityType.Other, hashedEmail);
+                if (hashedEmailMismatch && kitConfiguration != null && kitConfiguration.getHashedEmailUserIdentityType() != null) {
```
thomson-t, line 1556: Shouldn't we proceed with `identityBuilder.userIdentity(type, hashedEmail);` using `type = MParticle.IdentityType.Other` in the scenario where `getHashedEmailUserIdentityType` returns null?

jamesnrokt (approved):

android-kit-base/src/test/kotlin/com/mparticle/kits/KitManagerImplTest.kt
```diff
@@ -1105,10 +1105,11 @@ class KitManagerImplTest {
             String::class.java,
             MParticleUser::class.java,
             IdentityApi::class.java,
+            KitConfiguration::class.java,
             Runnable::class.java
         )
         method.isAccessible = true
-        val result = method.invoke(manager, "Test@gmail.com", "", user, identityApi, runnable)
+        val result = method.invoke(manager, "Test@gmail.com", "", user, identityApi, mockedKitConfig, runnable)
```
rmi22186, line 1112: I was writing this comment before we hopped on with Brandon, but it will still apply once the additional updates with an `Unassigned` value are included.

Can you add a test where the mockedkitconfig has a setting for `hashedEmailUserIdentityType` of something besides `other`?  Then the assertion would show that it makes a request with that identity?

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
```diff
@@ -995,6 +996,10 @@ public JSONArray getPlacementAttributesMapping() throws JSONException {
         return new JSONArray(jsonArrayStr);
     }
 
+    public String getHashedEmailUserIdentityType() {
+        return settings.get(KEY_HASHED_EMAIL_IDENTITY_TYPE);
```
BrandonStalnaker, line 1000: If this is already a getter maybe it should just go ahead and return the type from line 1561? might not work so not letting this question/suggestion block approval.

BrandonStalnaker (approved):
LGTM just had one comment

android-kit-base/src/main/java/com/mparticle/kits/KitConfiguration.java
Mansi-mParticle: KitConfiguration can be null, so we need this check for safety

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1523,13 +1523,23 @@ private void confirmEmail(
...
+            MParticle.IdentityType selectedIdentityType = null;
+            try {
+                String identityTypeStr = (kitConfiguration != null)
+                        ? kitConfiguration.getHashedEmailUserIdentityType()
+                        : null;
+                if (identityTypeStr != null && !identityTypeStr.equalsIgnoreCase("Unknown")) {
+                    selectedIdentityType = MParticle.IdentityType.valueOf(identityTypeStr);
+                }
+            } catch (IllegalArgumentException e) {
+                selectedIdentityType = MParticle.IdentityType.Other; // fallback if invalid value
```
rmi22186, line 1535: I don't think there should be a fallback to set this to a default. If there is no electedIdentityType, we shouldn't do any sort of identity request.

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1523,13 +1523,23 @@ private void confirmEmail(
         boolean hasHashedEmail = hashedEmail != null && !hashedEmail.isEmpty();
 
         if ((hasEmail || hasHashedEmail) && user != null) {
+            MParticle.IdentityType selectedIdentityType = null;
+            try {
+                String identityTypeStr = (kitConfiguration != null)
+                        ? kitConfiguration.getHashedEmailUserIdentityType()
+                        : null;
+                if (identityTypeStr != null && !identityTypeStr.equalsIgnoreCase("Unknown")) {
```
rmi22186, line 1531: I was planning to change this to an empty string, or "unselected" because that makes more sense.  Regardless of what it changes too, this method should be more generic and not require an actual string,  but rather grab an identity type based on a string of 'other, other2, other3" etc, and if it is 'unknown', or '', then return null

rmi22186: I think you can remove the check for `~identityTypeStr.equalsIgnoreCase("Unknown")` on line 1531, and just do 

```
selectedIdentityType = MParticle.IdentityType.valueOf(identityTypeStr);
} catch ... {
warn the user that the selected identity type isn't working.  
}
```

Then `selectedIdentityType` remains null from line 1526

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1523,13 +1523,23 @@ private void confirmEmail(
...
+            }
             String existingEmail = user.getUserIdentities().get(MParticle.IdentityType.Email);
-            String existingHashedEmail = user.getUserIdentities().get(MParticle.IdentityType.Other);
-
+            String existingHashedEmail = selectedIdentityType != null ? user.getUserIdentities().get(selectedIdentityType) : null;
             boolean emailMismatch = hasEmail && !email.equalsIgnoreCase(existingEmail);
             boolean hashedEmailMismatch = hasHashedEmail && !hashedEmail.equalsIgnoreCase(existingHashedEmail);
 
-            if (emailMismatch || hashedEmailMismatch) {
+            if (emailMismatch || (hashedEmailMismatch && selectedIdentityType != null)) {
```
rmi22186, line 1542: not sure if you will need `selectedIdentityType != null` anymore since it will be considered in order to get the `hashedEmailMismatch`, right?

android-kit-base/src/main/java/com/mparticle/kits/KitManagerImpl.java
```diff
@@ -1553,20 +1563,8 @@ else if (hashedEmailMismatch && existingHashedEmail != null) {
...
-                            type = MParticle.IdentityType.valueOf(identityTypeStr);
-                        }
-                    } catch (IllegalArgumentException e) {
-                        type = MParticle.IdentityType.Other; // fallback if invalid
-                    }
-
-                    if (type != null) {
-                        identityBuilder.userIdentity(type, hashedEmail);
-                    }
+                if (hashedEmailMismatch && selectedIdentityType != null) {
```
rmi22186, line 1566: Pretty sure `selectedIdentityType != null` can be removed.  If i am misunderstanding, please let me know

BrandonStalnaker (approved):

rmi22186 (approved):

PR #604, opened 2025-09-12
feat: Implement Trunk tool

Mansi-mParticle:
## Instructions

 ## Summary
 -  Added .trunk/trunk.yaml configuration for linting and formatting tools.

 ## Testing Plan
 

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -12,6 +12,18 @@ permissions:
   checks: write
   id-token: write
 
+  # Code quality checks
+  trunk-check:
+    name: Trunk code check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
+      - name: Trunk Check
+        uses: trunk-io/trunk-action@75699af9e26881e564e9d832ef7dc3af25ec031b # v1.2.4
+        with:
+          check-mode: all
+
 jobs:
   pr-check-hadcoded-secrets:
     name: "Check PR for hardcoded secrets"


.github/workflows/daily.yml
```diff
@@ -40,10 +45,10 @@ jobs:
       GIT_COMMITTER_EMAIL: developers@mparticle.com
     steps:
       - name: "Clone branch"
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: mparticle/mparticle-android-sdk
-          ref: ${{ inputs.branch_name }}
+          ref: developmentreq
```
thomson-t, line 51: Is this intentional? Do we need to hardcode this?

build.gradle
```diff
@@ -52,6 +52,18 @@ allprojects {
 
     apply plugin: 'org.jlleitschuh.gradle.ktlint'
     
+    ktlint {
+        // Disable problematic rules
+        disabledRules = [
+            "trailing-comma",
```
thomson-t, line 58: Should we disable this? I can see it auto-formatted most of the files.

build.gradle
```diff
@@ -52,6 +52,18 @@ allprojects {
...
+            "trailing-comma-on-call-site",
+            "trailing-comma-on-declaration-site",
+            "max-line-length",  // Disable line length checks
```
thomson-t, line 61: We can create a `.editorconfig` file with the required line length, and it should be automatically updated.

thomson-t (approved):
I noticed that the commit hash in `kits/rokt-kit` has changed. Is that intentional?

.github/workflows/pull-request.yml
```diff
@@ -222,7 +231,7 @@ jobs:
           submodules: recursive
           fetch-depth: 0
       - name: "Install JDK 17"
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
```
thomson-t: Can we add the trunk check as part of the PR workflow?
Something like 
```yml
      - name: Trunk Check
        uses: trunk-io/trunk-action@75699af9e26881e564e9d832ef7dc3af25ec031b # v1.2.4
        with:
          check-mode: all

```
Or we can do it as another PR as well.

thomson-t (approved):

PR #698, opened 2026-04-20
feat: Add max persistence age override option [TRIAGE-608]

junias-rokt:
## Background

The iOS mParticle SDK exposes `MParticleOptions.persistenceMaxAgeSeconds`, which caps the age of locally persisted events, batches, and sessions to prevent unbounded on-device storage growth. The Android SDK had no equivalent — once data was written to the SQLite store it would sit there until the app either uploaded it or was uninstalled. For apps that batch heavily, go long stretches offline, or see users who rarely background the app, this can produce noticeable storage bloat and a slow drift toward larger database files. TRIAGE-608 was opened to close this feature-parity gap and document it on the Android configuration guide.

The companion iOS implementation lives in [`MPBackendController.cleanUp:`](https://github.com/mParticle/mparticle-apple-sdk/blob/main/mParticle-Apple-SDK/MPBackendController.m) and [`MPPersistenceController deleteRecordsOlderThan:`](https://github.com/mParticle/mparticle-apple-sdk/blob/main/mParticle-Apple-SDK/Persistence/MPPersistenceController.mm), which is the behavioral spec this change mirrors.

## What Has Changed

**Public API**
- Adds `MParticleOptions.Builder#persistenceMaxAgeSeconds(int seconds)` and the matching getter `MParticleOptions#getPersistenceMaxAgeSeconds()`. Units are seconds, values must be positive; zero or negative values log a warning and fall back to the default.

**Wiring**
- Threads the value through `ConfigManager` so the upload pipeline can read it at runtime.

**Age-based sweep**
- New `MParticleDBManager#deleteRecordsOlderThan(long cutoffMillis)` orchestrator wraps three new static helpers in a single transaction:
  - `MessageService#deleteMessagesOlderThan` — deletes from `MessageTableColumns.TABLE_NAME` where `created_at < ?`
  - `UploadService#deleteUploadsOlderThan` — deletes from `UploadTableColumns.TABLE_NAME` where `created_at < ?`
  - `SessionService#deleteSessionsOlderThan` — deletes from `SessionTableColumns.TABLE_NAME` where `end_time < ?`
- `UploadHandler#upload()` now calls `maybePrunePersistedRecords(now)` at the start of each cycle. The sweep is throttled to run at most **once every 24 hours** and defaults to a **90-day** retention window when `persistenceMaxAgeSeconds` is unset — matching the iOS defaults precisely.

**Tests**
- `MParticleOptionsTest#testPersistenceMaxAgeSeconds` covers null/positive/zero/negative inputs on the builder.
- `MessageServiceTest#testDeleteMessagesOlderThan` verifies the SQL cutoff semantics (strictly `<` cutoff are deleted; rows at or newer than the cutoff are retained).

## Screenshots/Video

N/A — this is a configuration/storage-hygiene change with no UI surface.

## Checklist

- [x] I have made corresponding changes to the documentation.
- [x] I have tested this locally.

Local validation against JDK 17: \`trunk check\`, \`./gradlew build\`, \`./gradlew test\`, \`./gradlew ktlintCheck\`, and \`./gradlew lint\` all pass. \`connectedAndroidTest\` was not run locally (no emulator available) — will rely on CI.

## Additional Notes

- **Defaults are intentionally conservative.** With no opt-in, behavior changes from *never-delete* to *delete rows older than 90 days*, applied at most once per 24 hours. This matches long-standing iOS behavior and is the same floor SonarCloud/Bugbot flagged as low-risk on this PR.
- **Hook site.** The sweep runs inside the existing upload cycle rather than in a dedicated lifecycle hook. If apps are observed that never trigger an upload (e.g., very short-lived sessions with no events), a follow-up can additionally invoke \`maybePrunePersistedRecords\` from \`AppStateManager#onActivityPaused\`; the 24h throttle already coordinates across call sites so this is a safe additive change.
- **Docs.** Public documentation is updated in companion PR [mparticle-by-rokt/docsite#3390](https://github.com/mparticle-by-rokt/docsite/pull/3390).

## Reference Issue (For employees only. Ignore if you are an outside contributor)

- Closes https://rokt.atlassian.net/browse/TRIAGE-608

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticleOptions.java
+++ b/android-core/src/main/java/com/mparticle/MParticleOptions.java
@@ -42,6 +42,7 @@ public class MParticleOptions {
     private Integer mUploadInterval = ConfigManager.DEFAULT_UPLOAD_INTERVAL;  //seconds
     private Integer mSessionTimeout = ConfigManager.DEFAULT_SESSION_TIMEOUT_SECONDS; //seconds
     private Integer mConfigMaxAge = null;
+    private Integer mPersistenceMaxAgeSeconds = null;
     private Boolean mUnCaughtExceptionLogging = false;
     private MParticle.LogLevel mLogLevel = MParticle.LogLevel.DEBUG;
     private AttributionListener mAttributionListener;
@@ -116,6 +117,13 @@ public class MParticleOptions {
                 Logger.warning("Config Max Age must be a positive number, disregarding value.");
             } else {
                 this.mConfigMaxAge = builder.configMaxAge;
+            }
+        }
+        if (builder.persistenceMaxAgeSeconds != null) {
+            if (builder.persistenceMaxAgeSeconds <= 0) {
+                Logger.warning("Persistence Max Age must be a positive number, disregarding value.");
+            } else {
+                this.mPersistenceMaxAgeSeconds = builder.persistenceMaxAgeSeconds;
             }
         }
         if (builder.unCaughtExceptionLogging != null) {
@@ -290,6 +298,22 @@ public class MParticleOptions {
         return mConfigMaxAge;
     }
 
+    /**
+     * The maximum threshold (in seconds) for locally persisted events, batches, and sessions.
+     * <p>
+     * When {@code null} (the default), records are retained for 90 days before being deleted.
+     * Values less than or equal to zero are rejected at build time and result in the default
+     * being used.
+     *
+     * @return the configured maximum persistence age in seconds, or {@code null} when the default
+     *         (90 days) applies
+     * @see Builder#persistenceMaxAgeSeconds(int)
+     */
+    @Nullable
+    public Integer getPersistenceMaxAgeSeconds() {
+        return mPersistenceMaxAgeSeconds;
+    }
+
     @NonNull
     public Boolean isUncaughtExceptionLoggingEnabled() {
         return mUnCaughtExceptionLogging;
@@ -403,6 +427,7 @@ public class MParticleOptions {
         private Integer uploadInterval = null;
         private Integer sessionTimeout = null;
         private Integer configMaxAge = null;
+        private Integer persistenceMaxAgeSeconds = null;
         private Boolean unCaughtExceptionLogging = null;
         MParticle.LogLevel logLevel = null;
         BaseIdentityTask identityTask;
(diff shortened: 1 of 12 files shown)


android-core/src/main/java/com/mparticle/internal/database/services/MParticleDBManager.java
```diff
@@ -206,6 +206,31 @@ public void deleteMessagesAndSessions(String currentSessionId) {
...
+            MessageService.deleteMessagesOlderThan(db, cutoffMillis);
+            UploadService.deleteUploadsOlderThan(db, cutoffMillis);
+            SessionService.deleteSessionsOlderThan(db, cutoffMillis);
+            db.setTransactionSuccessful();
+        } catch (Exception e) {
+            Logger.warning(e, "Error pruning persisted records older than " + cutoffMillis + " ms.");
+        } finally {
+            db.endTransaction();
+        }
+    }
```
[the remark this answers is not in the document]
junias-rokt: This is correct and I ended up fixing it. The apple sdk displays the same behavior it looks like. Happy to file a ticket for it or reverse the fix here to preserve parity between the two sdks. Let me know your thoughts.

android-core/src/main/java/com/mparticle/MParticleOptions.java
```diff
@@ -118,6 +119,13 @@ public MParticleOptions(@NonNull Builder builder) {
                 this.mConfigMaxAge = builder.configMaxAge;
             }
         }
+        if (builder.persistenceMaxAgeSeconds != null) {
+            if (builder.persistenceMaxAgeSeconds <= 0) {
+                Logger.warning("Persistence Max Age must be a positive number, disregarding value.");
+            } else {
+                this.mPersistenceMaxAgeSeconds = builder.persistenceMaxAgeSeconds;
+            }
```
thomson-t, line 127: Shouldn't we check for the upper bound? Something like
```java
if (builder.persistenceMaxAgeSeconds <= 0) {
    Logger.warning("Persistence Max Age must be a positive number, disregarding value.");
} else if (builder.persistenceMaxAgeSeconds > MAX_PERSISTENCE_MAX_AGE_SECONDS) {
    Logger.warning("Persistence Max Age is too high");
    this.mPersistenceMaxAgeSeconds = MAX_PERSISTENCE_MAX_AGE_SECONDS;
} else {
    this.mPersistenceMaxAgeSeconds = builder.persistenceMaxAgeSeconds;
}
```
or not set it at all of not in the range?

junias-rokt: Hey good callout! This is intentional as the main aim for these changes was to have feature parity with the other SDKs. Changing this would diverge from the other ones. This also matches all the other validators in the file and large values basically just disable the sweep.

thomson-t: ok, fair enough

thomson-t (approved):

PR #700, opened 2026-05-13
refactor: migrate Rokt contracts and facade ownership to rokt-kit

denischilik:
## Background
This change completes the Rokt ownership migration to the `rokt-kit` module so Rokt-specific contracts no longer leak through `android-core` and `android-kit-base`.
It also removes fragile reflection in the Rokt enablement check to improve runtime reliability.

## What Has Changed
- Moved Rokt facade and related Rokt view/callback contracts into `kits/rokt/rokt`
- Replaced mParticle wrapper contracts with native `com.rokt.roktsdk` types across the Rokt kit flow
- Removed `RoktKitApi` bridging from core kit manager contracts
- Updated Rokt request and bridge flow to resolve and execute through kit-owned components only
- Reworked `isEnabled` resolution to use a callback provider instead of reflection and updated unit tests accordingly

## Screenshots/Video
N/A

## Checklist
- [x] I have tested this locally.

## Additional Notes
This PR is intentionally scoped to internal ownership/type migration for Rokt integration and related reliability fixes.


android-kit-base/src/main/java/com/mparticle/kits/KitIntegration.java
```diff
@@ -590,55 +582,16 @@ public interface BatchListener {
...
-
-        void enrichAttributes(
-                @NonNull Map<String, String> attributes, @Nullable FilteredMParticleUser user);
         /**
          * Set the SDK version of the mParticle SDK.
          * This should match the value set in MParticle.getWrapperSdkVersion()
          *
          * @param wrapperSdkVersion the version of the mParticle SDK
          */
         void setWrapperSdkVersion(@NonNull WrapperSdkVersion wrapperSdkVersion);
```
thomson-t, line 595: If this is used by just Rokt SDK, do we need to keep it?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Kotlin-friendly accessor for the legacy Rokt API object.
+ */
+fun MParticle.Rokt(): Rokt = createRokt(this)
+
+/**
+ * Java-friendly accessors for the legacy Rokt API object.
+ */
+object MParticleRokt {
+    @Suppress("FunctionName")
```
thomson-t, line 12: This is probably confusing. Can we keep the function name start with small letters?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,27 @@
...
+    @JvmStatic
```
thomson-t, line 13: Do we need `JvmStatic` here? Only consumer is Kotlin, isn't it?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRoktExtensions.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+fun MParticle.Rokt(): Rokt = createRokt(this)
+
+/**
+ * Java-friendly accessors for the legacy Rokt API object.
+ */
+object MParticleRokt {
+    @Suppress("FunctionName")
+    @JvmStatic
+    fun Rokt(mParticle: MParticle?): Rokt? = mParticle?.let { createRokt(it) }
```
thomson-t: Why do we need this nullable?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.rokt.roktsdk.Rokt.RoktCallback
+import com.rokt.roktsdk.RoktConfig
+import com.rokt.roktsdk.RoktEvent
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import java.lang.ref.WeakReference
+
+class Rokt internal constructor(private val mKitManager: KitManager) {
+    @JvmOverloads
+    fun selectPlacements(
```
thomson-t, line 30: Since this is a public api, it would be nice to have some javadoc documentation.

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    fun getSessionId(): String? = if (isEnabled()) {
+        resolveRoktKit()?.second?.getSessionId()
+    } else {
+        null
+    }
+
+    fun prepareAttributesAsync(attributes: Map<String, String>) {
```
thomson-t: We don't want this to be public.

android-kit-base/src/main/java/com/mparticle/kits/KitIntegration.java
denischilik: Good call. This is still part of the shared mParticle kit interface, and I didn’t want to expand this PR’s scope by changing that contract here.
Agreed it should be moved to the Rokt layer (kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt) in a follow-up PR and then removed from shared interfaces.

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
denischilik: Updated

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRoktExtensions.kt
denischilik: Updated

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt
denischilik: Make it internal thanks

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
denischilik: fixed

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt
denischilik: fixed

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        synchronized(this) {
+            rokt?.let { return it }
+
+            return createRokt(mParticle).also {
+                rokt = it
+            }
+        }
+    }
```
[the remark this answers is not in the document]
denischilik: The previous logic do the same, we don't need to update this

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Java-friendly accessors for the legacy Rokt API object.
+ */
+object MParticleRokt {
+    @Volatile
+    private var rokt: Rokt? = null
+
+    @Suppress("FunctionName")
+    @JvmStatic
+    fun Rokt(): Rokt {
```
jamesnrokt, line 14: This comes across slightly confusing as it's accessed like `MParticleRokt.Rokt()` - How about `MParticleRokt.getInstance()` or similar?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,32 @@
+package com.mparticle.kits
+
+import com.mparticle.MParticle
+
+/**
+ * Java-friendly accessors for the legacy Rokt API object.
+ */
+object MParticleRokt {
+    @Volatile
```
jamesnrokt, line 9: Do we need this? rokt is only read inside synchronized(this) at line 20

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
jamesnrokt: Agree, part of my comment below is if it should be called Rokt at all though

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+object MParticleRokt {
+    @Suppress("FunctionName")
+    @JvmStatic
+    fun Rokt(): Rokt {
+        val mParticle = requireNotNull(MParticle.getInstance()) {
+            "MParticle must be started before calling MParticleRokt.Rokt()"
+        }
+        return createRokt(mParticle)
+    }
```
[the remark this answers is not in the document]
jamesnrokt: This seems legit as we used to cache it in the mParticle class?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/Rokt.kt
jamesnrokt: Agree, any reason not to do this?

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
denischilik: We access it with MParticle.getInstance().Rokt() like before. That is extension to MParticle class

BrandonStalnaker (approved):
LGTM

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
denischilik: In this code, @Volatile is used to ensure correct lazy singleton initialization in a multithreaded environment.

kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/MParticleRokt.kt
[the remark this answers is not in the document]
denischilik: Yep, we cashed it before in mparticle too

PR #701, opened 2026-05-13
feat: add customBaseURL CNAME support to NetworkOptions

jamesnrokt:
## Summary

Adds `NetworkOptions.Builder.setCustomBaseURL(String)` which routes all mParticle endpoint traffic (config, events, identity, alias, audience) through a single HTTPS CNAME host. Mirrors iOS work from [mparticle-apple-sdk#760](https://github.com/mParticle/mparticle-apple-sdk/pull/760).

## What changed

- **`NetworkOptions`**: new `customBaseURL` field + getter; `Builder.setCustomBaseURL(@NonNull String)` validates HTTPS scheme, requires a host, and stores host(:port) only — path/query/fragment ignored.
- **`MParticleBaseClientImpl.getUrl`**: when `customBaseURL` is set:
  - Skips the pod prefix (the CNAME's CDN handles regional routing).
  - Logs a warning and ignores any per-endpoint `DomainMapping` and `overridesSubdirectory` (unsupported with CDN routing).
  - Rewrites paths to: `/config/v4/<key>/config`, `/nativeevents/v2/<key>/events`, `/identity/v1/<path>`, `/nativeevents/v1/identity/<key>/alias`, `/nativeevents/v1/<key>/audience`.
- **`android-core/proguard.pro`**: adds `-keep` rules for `MParticle$Internal` and `ConfigManager.getNetworkOptions()` so downstream kits can navigate the public extension surface after R8 minification.

## Downstream

The Rokt kit reads `NetworkOptions.customBaseURL` and forwards it to the Rokt SDK: https://github.com/mparticle-integrations/mparticle-android-integration-rokt/pull/143

## Test plan

Added 9 androidTest cases in `MParticleBaseClientImplTest`:
- [x] Custom CNAME routes config endpoint via `/config/v4/`
- [x] Custom CNAME routes events endpoint via `/nativeevents/v2/`
- [x] Custom CNAME routes identity endpoint via `/identity/v1/`
- [x] Custom CNAME routes alias endpoint via `/nativeevents/v1/identity/`
- [x] Custom CNAME routes audience endpoint via `/nativeevents/v1/<key>/audience`
- [x] Non-HTTPS values rejected
- [x] Path/query stripped
- [x] Port preserved
- [x] Malformed URL rejected

Locally: `:android-core:testDebugUnitTest` passes (existing unit suite); `:android-core:compileDebugAndroidTestKotlin` compiles. AndroidTest suite runs on CI emulator.

### DIFF
--- a/android-core/proguard.pro
+++ b/android-core/proguard.pro
@@ -84,6 +84,10 @@
 
 -keep class com.mparticle.MPEvent$* { *; }
 -keep class com.mparticle.MParticle { *; }
+-keep class com.mparticle.MParticle$Internal { *; }
+-keep class com.mparticle.internal.ConfigManager {
+    public com.mparticle.networking.NetworkOptions getNetworkOptions();
+}
 -keep class com.mparticle.MParticle$EventType { *; }
 -keep class com.mparticle.MParticle$InstallType { *; }
 -keep class com.mparticle.MParticle$IdentityType { *; }
--- a/android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
+++ b/android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
@@ -16,11 +16,13 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.HashSet;
 import java.util.Set;
 
 public class NetworkOptions {
@@ -66,6 +68,13 @@ public class NetworkOptions {
             JSONObject jsonObject = new JSONObject(jsonString);
             builder.setPinningDisabledInDevelopment(jsonObject.optBoolean("disableDevPinning", false));
             builder.setPinningDisabled(jsonObject.optBoolean("disablePinning", false));
+            String storedCustomBaseURL = jsonObject.optString("customBaseURL", null);
+            if (!MPUtility.isEmpty(storedCustomBaseURL)) {
+                // Stored value is already host(:port). Skip the Builder setter (which expects a full
+                // HTTPS URL with scheme) and assign directly so a previously-validated value survives
+                // the JSON round-trip.
+                builder.customBaseURL = storedCustomBaseURL;
+            }
             JSONArray domainMappingsJson = jsonObject.getJSONArray("domainMappings");
             for (int i = 0; i < domainMappingsJson.length(); i++) {
                 builder.addDomainMapping(DomainMapping
@@ -138,6 +147,9 @@ public class NetworkOptions {
             JSONArray domainMappingsJson = new JSONArray();
             networkOptions.put("disableDevPinning", pinningDisabledInDevelopment);
             networkOptions.put("disablePinning", pinningDisabled);
+            if (!MPUtility.isEmpty(customBaseURL)) {
+                networkOptions.put("customBaseURL", customBaseURL);
+            }
             networkOptions.put("domainMappings", domainMappingsJson);
             for (DomainMapping domainMapping : domainMappings.values()) {
                 domainMappingsJson.put(domainMapping.toString());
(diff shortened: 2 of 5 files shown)


android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
```diff
@@ -187,6 +203,40 @@ public Builder setPinningDisabled(boolean disabled) {
...
+         * <p>Certificate pinning: if pinning is enabled (default), supply certificates for
+         * the CNAME domain via the relevant {@link DomainMapping}, or disable pinning via
+         * {@link #setPinningDisabled(boolean)} / {@link #setPinningDisabledInDevelopment(boolean)}.
+         *
+         * @param customBaseURL HTTPS URL containing the CNAME host
+         */
+        @NonNull
+        public Builder setCustomBaseURL(@NonNull String customBaseURL) {
+            try {
+                java.net.URL parsed = new java.net.URL(customBaseURL);
```
thomson-t: nit: import the class.

android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
```diff
@@ -187,6 +203,40 @@ public Builder setPinningDisabled(boolean disabled) {
...
+                if (!"https".equalsIgnoreCase(parsed.getProtocol()) || MPUtility.isEmpty(parsed.getHost())) {
+                    Logger.warning("NetworkOptions: customBaseURL must use HTTPS and include a valid host — value ignored.");
+                    return this;
+                }
+                StringBuilder host = new StringBuilder(parsed.getHost());
+                if (parsed.getPort() != -1) {
+                    host.append(":").append(parsed.getPort());
+                }
+                this.customBaseURL = host.toString();
+            } catch (java.net.MalformedURLException e) {
```
thomson-t: nit: import the class here as well.

android-core/proguard.pro
```diff
@@ -84,6 +84,10 @@
 
 -keep class com.mparticle.MPEvent$* { *; }
 -keep class com.mparticle.MParticle { *; }
+-keep class com.mparticle.MParticle$Internal { *; }
+-keep class com.mparticle.internal.ConfigManager {
+    public com.mparticle.networking.NetworkOptions getNetworkOptions();
+}
```
thomson-t, line 90: Good catch

thomson-t (dismissed):

denischilik (commented):
Looks good to me but I think `customBaseURL lost during database serialization round-trip` should be addressed

android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
```diff
@@ -137,6 +152,7 @@ public static class Builder {
         private Map<Endpoint, DomainMapping> domainMappings = new HashMap<Endpoint, DomainMapping>();
         private Boolean pinningDisabledInDevelopment;
         private Boolean pinningDisabled;
+        private String customBaseURL;
```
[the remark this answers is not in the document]
jamesnrokt: Good catch — fixed in 42fff2a2. `toJson()` now writes `customBaseURL` (when non-empty) and `withNetworkOptions()` reads it back. The stored value is already host(:port) so I assign it directly on the builder instead of going through the HTTPS-validating setter. Added `testCustomBaseURLSurvivesJsonRoundTrip` and `testCustomBaseURLOmittedFromJsonWhenUnset` to lock in the behavior.

android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
jamesnrokt: Done — imported `java.net.URL` and `java.net.MalformedURLException` in 42fff2a2.

android-core/src/main/java/com/mparticle/networking/NetworkOptions.java
jamesnrokt: Done in 42fff2a2.

thomson-t (dismissed):

thomson-t (approved):

PR #703, opened 2026-05-18
Handle legacy persisted domain mappings safely

ankitsingh08:
## Background
- This issue appears to have been introduced after [#588](https://github.com/mParticle/mparticle-android-sdk/pull/588), where persisted `DomainMapping` parsing began expecting `overridesSubdirectory` to always be present.
- Users upgrading from older Android SDK versions can still have persisted `UploadSettings` / `NetworkOptions` JSON containing legacy `domainMappings` entries that do not include `overridesSubdirectory`.
- In mParticle logs, this surfaced as:
  - `No value for overridesSubdirectory`
  - `Attempt to invoke virtual method 'com.mparticle.networking.DomainMapping com.mparticle.networking.DomainMapping$Builder.build()' on a null object reference`
- The current parser expected `overridesSubdirectory` to always be present, which caused legacy entries to fail parsing.
- In the persisted `NetworkOptions` flow, that parse failure could produce a null `DomainMapping.Builder`, and the subsequent `.build()` call could trigger a null object reference crash.
- Because these values are read from persisted upload settings, affected users could continue seeing repeated failures until the stale rows were replaced.

## What Has Changed
- Updated `DomainMapping` JSON parsing to use a backward-compatible default of `false` when `overridesSubdirectory` is missing.
- Hardened `NetworkOptions` parsing so invalid persisted domain mapping entries are skipped instead of crashing the entire parse flow.
- Added regression tests covering:
  - legacy persisted `DomainMapping` JSON without `overridesSubdirectory`
  - `NetworkOptions` JSON with one valid persisted mapping and one invalid mapping

## Screenshots/Video
- Not applicable for this change.

## Checklist
- [ ] I have tested this locally.

## Additional Notes
- Local validation run:
  - `./gradlew :android-core:testDebugUnitTest --tests com.mparticle.networking.NetworkOptionsTest`
- The fix is intentionally minimal and additive to preserve backward compatibility and avoid API changes.

### DIFF
--- a/android-core/src/main/java/com/mparticle/networking/DomainMapping.java
+++ b/android-core/src/main/java/com/mparticle/networking/DomainMapping.java
@@ -326,7 +326,7 @@ public class DomainMapping {
                 JSONObject jsonObject = new JSONObject(jsonString);
                 int type = jsonObject.getInt("mType");
                 String newUrl = jsonObject.getString("url");
-                boolean overridesSubdirectory = jsonObject.getBoolean("overridesSubdirectory");
+                boolean overridesSubdirectory = jsonObject.optBoolean("overridesSubdirectory", false);
                 Builder builder = new Builder(Endpoint.parseInt(type), newUrl, overridesSubdirectory);
                 JSONArray certificatesJsonArray = jsonObject.getJSONArray("mCertificates");
                 for (int i = 0; i < certificatesJsonArray.length(); i++) {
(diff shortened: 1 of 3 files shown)


jamesnrokt (approved):
Verified locally against the Higgs Shop sample app on Pixel_9 emulator.

Pass A (main): legacy JSON → JSONException + NPE matching the report, domainMappings.size=0 (all mappings silently dropped).
Pass B (this branch): same input → domainMappings.size=2, configDomain populated with overridesSubdirectory=false default, no exceptions.

Reproduces the original failure and confirms the fix. LGTM.

Thanks for the clean, minimal fix and the regression tests — great first contribution, looking forward to more!

PR #711, opened 2026-05-26
feat(rokt): add kit API parity access

thomson-t:
## Background

The Rokt kit migration for 6.0 moved partner-facing Rokt APIs out of the core SDK, but Kotlin consumers still need a low-friction way to access the kit facade from the active mParticle instance. The Rokt SDK Compose wrapper also exposes native placement events that were not forwarded through the mParticle kit wrapper.

## What Has Changed

- Add Kotlin `MParticle.rokt` extension access for the Rokt kit facade.
- Keep the Java-friendly `MParticleRokt.Rokt()` helper as the supported Java entry point.
- Forward `RoktLayout` `onEvent` callbacks to the native Rokt SDK composable.
- Add tests for the Kotlin extension facade and stale instance refresh behavior.
- Update the Rokt kit README and changelog with the restored API shape.

## Screenshots/Video

N/A - no visual changes.

## Checklist

- [x] Self-review completed
- [x] Tests added or updated
- [x] Tested locally

## Validation

- `trunk check`
- `trunk check --ci`
- `git diff --check`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home ./gradlew build`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home ./gradlew test`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home ./gradlew ktlintCheck`
- `cd kits/rokt/rokt && ./gradlew build -Dorg.gradle.java.home=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home`
- `cd kits/rokt/rokt && ./gradlew test -Dorg.gradle.java.home=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home`
- `cd kits/rokt/rokt && ./gradlew ktlintCheck -Dorg.gradle.java.home=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home`

PR #712, opened 2026-05-27
ci: remove cross-platform-tests and semantic PR/branch checks

jamesnrokt:
## Summary
Removes three checks that have been failing on nearly every recent PR (#690–#711):

- `cross-platform-tests` — entire workflow deleted (known broken; depends on the external `mParticle/crossplatform-sdk-tests` repo)
- `Check PR for semantic title` — `pr-title-check` job removed from `pull-request.yml`
- `Check PR for semantic branch name` — `pr-branch-check-name` job removed from `pull-request.yml`

## Notes
- If any of these were listed as required checks in branch protection for `main`, those rules will need to be updated in repo settings so new PRs don't get stuck waiting on checks that no longer run.
- Two other repeat-offender failures were found but **not** removed here:
  - `Notify GChat / notify-gchat` — fails on fork PRs / missing secret; it's a notification, not a gate. Worth a follow-up to add `continue-on-error` or remove.
  - `SonarCloud Code Analysis` — posted by the SonarCloud GitHub App, not a workflow file. Needs to be turned off in the SonarCloud UI / repo app settings.

## Test plan
- [ ] CI on this PR shows the three removed checks no longer appear
- [ ] Remaining checks (unit tests, lint, instrumented tests, kit compatibility, security, trunk) still run


BrandonStalnaker:
Totally agree with removing the cross-platform-tests but do we no longer want standardized branch names and commit messages?

jamesnrokt:
> Totally agree with removing the cross-platform-tests but do we no longer want standardized branch names and commit messages?

For PR titles I felt that these are usually superfluous as they usually fail on the first opening PR and then can never pass so it was adding noise with ❌ on all the commits. Given we no longer use the titles for semver it seems like a fairly useless check and something we can easily edit before merge.

For branch names I don't see the value at all as our branch protection rules protect us from issues and it doesn't impact the work itself.

Wondering what your thoughts are given the above? Overall I'm happy to change it if it causes an issue, my main goal is at a glance ✅ on commits

nickolas-dimitrakas:
> > Totally agree with removing the cross-platform-tests but do we no longer want standardized branch names and commit messages?
> 
> For PR titles I felt that these are usually superfluous as they usually fail on the first opening PR and then can never pass so it was adding noise with ❌ on all the commits. Given we no longer use the titles for semver it seems like a fairly useless check and something we can easily edit before merge.
> 
> For branch names I don't see the value at all as our branch protection rules protect us from issues and it doesn't impact the work itself.
> 
> Wondering what your thoughts are given the above? Overall I'm happy to change it if it causes an issue, my main goal is at a glance ✅ on commits

I second the bit about not being able to correct the PR titles for the ✅. It was tedious to have to open a new PR with a proper branch name. No problem with any of the other jobs that were removed either.

BrandonStalnaker (approved):

thomson-t (approved):

PR #728, opened 2026-07-06
feat: Support Braze 43 and recommended eCommerce events

nickolas-dimitrakas:
## Background

Braze Android SDK 40.3.0+ introduces [recommended eCommerce events](https://www.braze.com/docs/developer_guide/analytics/logging_ecommerce_events/#android) (ecommerce.cart_updated, ecommerce.checkout_started, ecommerce.product_viewed, ecommerce.order_placed, ecommerce.order_refunded) as typed SDK APIs that unlock calculated profile fields, eCommerce reporting, and out-of-the-box Canvas templates. Braze 43.0.0 adds optional typed subtotalValue, tax, and shipping fields on CartUpdatedEvent, CheckoutStartedEvent, and OrderPlacedEvent. This mirrors the iOS work in 
[mparticle-apple-sdk#793](https://github.com/mParticle/mparticle-apple-sdk/pull/793).

## What changed

- New braze-43 kit track targeting com.braze:android-sdk-ui:[43.0.0,44.0.0) (namespace appboy43, artifact com.mparticle:braze-43), evolved from braze-41 (via an intermediate braze-42 track that was renamed to align with Braze 43).
- Braze 43 requires Kotlin 2.2.x, so the track is built standalone following the urbanairship-20 convention: excluded from settings-kits.gradle, its own Kotlin 2.2.20 buildscript, and dedicated CI steps in pull-request.yml / daily.yml. Documented in ONBOARDING.md.
- New opt-in useEcommerceRecommendedEvents setting. When enabled, the six supported mParticle commerce actions map to Braze's recommended events via logEcommerceEvent (and logCustomEvent for ecommerce.order_refunded, which has no typed API):

  | mParticle action | Braze event |
  | :-- | :-- |
  | add_to_cart / remove_from_cart | `ecommerce.cart_updated` (action add/remove) |
  | checkout | `ecommerce.checkout_started` |
  | view_detail | `ecommerce.product_viewed` (one per product) |
  | purchase | `ecommerce.order_placed` |
  | refund | `ecommerce.order_refunded` |

- When the setting is off (default), commerce forwarding is unchanged. Unsupported actions and events without products fall back to legacy forwarding.
- Braze 43 typed field mapping:
- TransactionAttributes.tax → typed tax
- TransactionAttributes.shipping → typed shipping
- commerce custom attribute subtotal_value → typed subtotalValue
- Same values are also sent as top-level properties on ecommerce.order_refunded custom events
- Promoted fields are excluded from metadata to avoid duplication
[...]

PR #733, opened 2026-07-16
fix(braze-42): honor replaceSkuWithProductName on recommended eCommerce path

nickolas-dimitrakas:
## Summary
Follow-up to #728 addressing [Cursor Bugbot's "SKU replacement flag ignored" finding](https://github.com/mParticle/mparticle-android-sdk/pull/728#discussion_r3532420188).

- Recommended eCommerce forwarding (`useEcommerceRecommendedEvents`) always used `product.sku` as `productId`, while legacy `logPurchase` honors `replaceSkuWithProductName`. The same commerce event could therefore identify products differently in Braze depending on which forwarding path ran.
- Adds `recommendedProductId()`, mirroring the legacy flag check, and uses it everywhere the recommended path emits a product identifier: line items (`recommendedLineItems`), `ProductViewedEvent`, and the `order_refunded` JSON payload (`recommendedProductsJson`).
- Also removes a redundant explicit `it ->` lambda label in `logEvent` flagged in [review](https://github.com/mParticle/mparticle-android-sdk/pull/728#discussion_r3572121933).

Stacked on #728 — this branch is based on `feat/braze-42-recommended-ecommerce` and should be retargeted to `workstation/6.0-Release` once #728 merges.

## Test plan
- [x] `RecommendedEcommerceTests` — added coverage for the flag on the cart/line-item, product-detail, and refund paths (12/12 tests pass)
- [x] `ktlintCheck` passes on the kit

### DIFF
--- a/kits/braze/braze-42/src/main/kotlin/com/mparticle/kits/AppboyKit.kt
+++ b/kits/braze/braze-42/src/main/kotlin/com/mparticle/kits/AppboyKit.kt
@@ -190,7 +190,7 @@ open class AppboyKit :
         } else {
             val properties = BrazeProperties()
             val brazePropertiesSetter = BrazePropertiesSetter(properties, enableTypeDetection)
-            event.customAttributeStrings?.let { it ->
+            event.customAttributeStrings?.let {
                 for ((key, value) in it) {
                     newAttributes[key] = brazePropertiesSetter.parseValue(key, value)
                 }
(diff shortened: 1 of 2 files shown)


BrandonStalnaker:
This does not apply to the new path

denischilik:
@nickolas-dimitrakas mention that this is not required by Braze, so PR can be closed.

PR #739, opened 2026-07-21
chore: exclude adobemedia-3 from 6.0 publish and enable manual publish trigger

denischilik:
## Background

The 6.0.0 publish uploaded the `com.mparticle` kits as a single Central Portal deployment bundle. One component, `com.mparticle:adobemedia-3:6.0.0`, failed validation:

> Dependency version information is missing for dependency: com.adobe.marketing.mobile:core (and the other Adobe deps)

`adobemedia-3` resolves its Adobe dependencies through a BOM (`platform('com.adobe.marketing.mobile:sdk-bom:[3.0.0,4.0.0)')`) with versionless `api` declarations, so the generated POM omits the dependency versions, which Maven Central rejects. Because a Central Portal deployment is atomic, this single failure blocks the whole kits bundle (including `android-rokt-kit`, which `com.rokt:rokt-sdk-plus:6.0.0` depends on).

## What changed

- Exclude `:kits:adobemedia:adobemedia-3` from `settings-kits.gradle` until its Adobe dependency versions are pinned in the published POM. It will be reintroduced in a follow-up once fixed.
- Add `workflow_dispatch` to `Release – Publish` so the 6.0.0 kits bundle can be re-published without changing `VERSION` (the version must remain 6.0.0 since the already-published umbrella references `com.mparticle:*:6.0.0`).

## Test plan

- [ ] Merge, then run `Release – Publish` via workflow_dispatch on `main`.
- [ ] Confirm the kits deployment validates (43/43) in Central Portal and publishes.
- [ ] Confirm `com.mparticle:android-rokt-kit:6.0.0` and core artifacts resolve on Maven Central, unblocking `com.rokt:rokt-sdk-plus:6.0.0`.


.github/workflows/release-publish.yml
```diff
@@ -1,6 +1,7 @@
 name: Release – Publish
 
 on:
+  workflow_dispatch:
```
jamesnrokt: We shouldn't have workflow dispatch on releases as the flow is controlled by file as the trigger

jamesnrokt (commented):
Think the cursor bot has also called out a valid issue?

denischilik:
Addressed both pieces of feedback:

- **@jamesnrokt** — reverted the `workflow_dispatch` trigger; `release-publish.yml` is now byte-identical to `main`, so the release flow stays controlled solely by the `VERSION` file trigger. To re-publish 6.0.0 with `adobemedia-3` excluded (version must stay 6.0.0 since the already-published `com.rokt:rokt-sdk-plus:6.0.0` references `com.mparticle:*:6.0.0`), this PR touches `VERSION` (still `6.0.0` on line 1) so merging re-runs `Release – Publish` via the normal push trigger. Re-uploading the same version is safe — Central Portal creates a fresh deployment and the gradle publish task succeeds (already observed with the `rokt-sdk-plus` re-upload); the duplicate `core`/`rokt-sdk-plus` deployments are simply dropped, and the new 43-kit bundle publishes.

- **Bugbot** — `kits/matrix.json` and `settings-kit-examples.gradle` are now in sync with the `settings-kits.gradle` exclusion (adobemedia-3 removed from all three), so the Build Kits matrix no longer targets a missing project.

`adobemedia-3` will return in a follow-up once its Adobe BOM dependency versions are pinned in the published POM.

jamesnrokt (approved):

.github/workflows/release-publish.yml
denischilik: Removed

PR #753, opened 2026-08-06
fix(publishing): restore Maven Central publishing for isolated kits

thomson-t:
## Background

The 6.0.1 release workflow attempted to publish four Kotlin-isolated kits from
their standalone Gradle roots. Those builds could not access the root-only
`buildSrc` publication convention, so the Maven Central publication task was
not registered and the isolated artifacts were not published.

## What Has Changed

- Move the Maven publication convention from `buildSrc` into shared
  `build-logic` that is available to root, aggregate-kit, and standalone-kit
  builds.
- Configure stable artifact IDs and exact release-version resolution for
  `urbanairship-20`, `braze-43`, `ga-23`, and `ga4-23`.
- Validate generated POMs for Maven Central publication and run Maven-local
  smoke tests before uploading release artifacts.
- Update the draft and publish workflows to validate every isolated kit before
  Maven Central uploads.
- Document the isolated-kit coordinates, development workflow, and
  post-release verification process.

## Screenshots/Video

N/A — build and release workflow changes only.

## Checklist

- [x] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


.github/workflows/release-publish.yml
```diff
@@ -65,16 +82,16 @@ jobs:
...
+        run: ./gradlew -Pmparticle.kit.mparticleFromMavenLocalOnly=true -p kits/braze/braze-43 -PVERSION=${{ needs.setup-and-version.outputs.final_version }} publishMavenPublicationToMavenCentralRepository --no-build-cache
 
       - name: Publish Isolated Kits (ga-23)
-        run: ./gradlew -p kits/ga/ga-23 -PisRelease=true -PVERSION=${{ needs.setup-and-version.outputs.final_version }} publishMavenPublicationToMavenCentralRepository --no-build-cache
+        run: ./gradlew -Pmparticle.kit.mparticleFromMavenLocalOnly=true -p kits/ga/ga-23 -PVERSION=${{ needs.setup-and-version.outputs.final_version }} publishMavenPublicationToMavenCentralRepository --no-build-cache
 
       - name: Publish Isolated Kits (ga4-23)
-        run: ./gradlew -p kits/ga4/ga4-23 -PisRelease=true -PVERSION=${{ needs.setup-and-version.outputs.final_version }} publishMavenPublicationToMavenCentralRepository --no-build-cache
+        run: ./gradlew -Pmparticle.kit.mparticleFromMavenLocalOnly=true -p kits/ga4/ga4-23 -PVERSION=${{ needs.setup-and-version.outputs.final_version }} publishMavenPublicationToMavenCentralRepository --no-build-cache
```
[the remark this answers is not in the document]
denischilik: I think should be fixed

.github/workflows/release-publish.yml
```diff
@@ -1,6 +1,7 @@
 name: Release – Publish
 
 on:
+  workflow_dispatch:
```
denischilik: Do we need to add workflow_dispatch or it is temporary and will be remove after release?

BrandonStalnaker: Is this so we can run a failed release without doing a new one @thomson-t ?

thomson-t: Kept it as a recovery mechanism if in case the isolated kit publishing fails. Since we have the final step in mavenCentral, we can drop the partial published versions and rerun it. Happy to remove it if everything goes well.

.github/workflows/release-publish.yml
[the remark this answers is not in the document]
thomson-t: Publish to mavenCentral won't make it immediately available. We can drop it if it is a partial publish. 
But I will remove the workflow dispatch as it makes the flow inconsistent.

.github/workflows/release-publish.yml
thomson-t: Updated

denischilik (approved):

BrandonStalnaker (approved):

PR #759, opened 2026-08-18
feat: report mParticle API usage through the Rokt kit

thomson-t:
## Background

Rokt diagnostics can identify native SDK API usage, but calls made through mParticle are not currently distinguishable. This adds bounded, non-sensitive API identifiers only when the Rokt kit is active. Android port of mParticle/mparticle-apple-sdk#813; requires Rokt SDK `6.0.3`, which exposes the public `Rokt.logMParticleApiCall` entry point.

## What Has Changed

- Instrument public mParticle core, identity, user, session, messaging, media, and Rokt-facade APIs to forward a fixed, uppercase code vocabulary through the **active** Rokt kit only (no-op otherwise), via a decoupled `RoktApiDiagnosticsForwarder` interface so `android-core` stays free of kit types.
- **Reported APIs match iOS #813**: dropped identifiers iOS deliberately excludes (single-attribute `setUserAttribute`) and Android-only internal machinery that would misreport SDK plumbing as partner calls (identity getters, identity-state listeners); renamed `SET_UPDATE_INTERVAL` → `SET_UPLOAD_INTERVAL`; added `RESET`, `SET_LOG_LEVEL`, `GET_INTEGRATION_ATTRIBUTES`, `GET_ATTRIBUTION_INFO`, `IS_KIT_ACTIVE`, `GET_KIT_INSTANCE`.
- Kept Android-specific public APIs that are neither internal nor noisy (install referrer, non-auto push toggles, media audio state); dropped internally/auto-fired ones (location tracking, uncaught-exception toggles, push registration, `enablePushNotifications`).
- The mParticle **Rokt facade** (`kits/rokt/rokt/.../Rokt.kt`) reports a distinct `[MP_API_*]` set (`SELECT_PLACEMENTS`, `SELECT_SHOPPABLE_ADS`, `PURCHASE_FINALIZED`, `ROKT_EVENTS`, `ROKT_CLOSE`, `ROKT_SET_SESSION_ID`, `ROKT_GET_SESSION_ID`, `REGISTER_PAYMENT_EXTENSION`) — additive by design with the native SDK's own `[API_*]` diagnostics, so drop-offs between the facade call and the native call are observable.
- **Never over-log**: diagnostics are suppressed while SDK/kit internals invoke public APIs — auto-upload, `logLtvIncrease` → `logEvent`, push-token `modify`, and the Rokt kit's attribute enrichment / event + identity forwarding — via `MParticle.withoutRoktApiUsage`.
- **No delay / no crash / no retry**: forwarding is a bounded synchronous lookup on the caller thread; the code is validated and buffered inside the Rokt SDK (`logMParticleApiCall`). API arguments and values are never included — code identifiers only.

## Testing

- `:android-core:testDebugUnitTest` for `MParticleTest` and `KitFrameworkWrapperTest` pass locally, including new tests for active-kit gating and internal-call suppression.
- Note: the Rokt-kit module build could not be run in the authoring sandbox (Maven Central `429` rate-limiting on `com.rokt:roktsdk:6.0.3`); it will be validated in CI.

## Checklist

- [x] Self-review completed
- [x] Tests added or updated
- [ ] Tested locally (core: yes; rokt kit: pending CI — see note)

### DIFF
--- a/android-core/src/main/java/com/mparticle/MParticle.java
+++ b/android-core/src/main/java/com/mparticle/MParticle.java
@@ -148,6 +148,7 @@ public class MParticle {
      * @param interval in seconds
      */
     public void setUpdateInterval(int interval) {
+        MParticle.logRoktApiUsage("SET_UPLOAD_INTERVAL");
         long intervalMillis = interval * 1000L;
         if ((intervalMillis >= 1 && mConfigManager.getUploadInterval() != intervalMillis)) {
             upload();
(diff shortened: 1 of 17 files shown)


thomson-t:
Superseded by #761 so the head branch lives directly in mParticle/mparticle-android-sdk rather than a fork. The replacement uses the same reviewed commits and preserves the SDK Team review request.

PR #760, opened 2026-08-18
feat(rokt): add Rokt session handoff APIs

jaissica12:
## Background

Rokt Android SDK 6.1.1 provides supported token-backed session handoff for native and non-native integrations. Exposing this capability through the mParticle Rokt facade lets apps preserve authenticated Rokt continuity across native and WebView experiences.

## What Has Changed

- Add a public `RoktSession` with required session ID and token values plus optional expiry.
- Add `setSession` and `getSession` APIs to the Rokt facade and delegate to the supported Rokt SDK session APIs.
- Reject blank session IDs or tokens without falling back to the deprecated ID-only API.
- Retain deprecated `setSessionId` and `getSessionId` APIs for released-client compatibility.
- Document Kotlin and Java handoff usage, expiry behavior, and add unit coverage for valid, invalid, disabled, missing, and absent-session cases.

## Screenshots/Video

N/A — no visual changes.

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have tested this locally.

## Additional Notes

- Rokt Android SDK and payment extension 6.1.1 are inherited from `main`; this PR includes no dependency-version changes.

### DIFF
--- a/kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/RoktKit.kt
+++ b/kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/RoktKit.kt
@@ -19,6 +19,7 @@ import com.mparticle.kits.KitIntegration.CommerceListener
 import com.mparticle.kits.KitIntegration.IdentityListener
 import com.mparticle.kits.KitIntegration.RoktListener
 import com.mparticle.rokt.RoktApiDiagnosticsForwarder
+import com.mparticle.rokt.RoktSession
 import com.rokt.roktsdk.PlacementOptions
 import com.rokt.roktsdk.Rokt
 import com.rokt.roktsdk.Rokt.SdkFrameworkType.Android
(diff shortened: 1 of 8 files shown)


kits/rokt/rokt/src/main/kotlin/com/mparticle/kits/RoktKit.kt
```diff
@@ -337,22 +339,67 @@ class RoktKit :
...
+        if (sessionToken.isNotEmpty()) {
+            Rokt.setSession(
+                NativeRoktSession(
+                    sessionId = sessionId,
+                    sessionToken = sessionToken,
+                    expiresAt = session.expiresAt,
+                ),
+            )
+        } else {
+            Rokt.setSessionId(sessionId)
```
thomson-t, line 365: Do we still need this? 
The new SDK won't be using this anyway.

BrandonStalnaker: We've only deprecated it for now but I could see removing if you want.

thomson-t (approved):

PR #770, opened 2026-09-02
ci: track Java to Kotlin migration progress on every PR

nickolas-dimitrakas:
First PR of the Java → Kotlin migration. It builds the instrument before we start
cutting: a per-PR progress report, and a ratchet that stops the problem growing.

Plan and work breakdown: [`docs/kotlin-migration.md`](docs/kotlin-migration.md).

## What this adds

**`scripts/kotlin-migration-progress.sh`** — measures Kotlin vs Java LOC across the three
in-scope `src/main` source sets (`android-core`, `android-kit-base`, `kits/rokt/rokt`),
excluding `build/` output, test source sets and sample apps. Emits a markdown report or
`KEY=value` metrics.

**`scripts/kotlin-migration-facades.txt`** — the reviewed list of Java files that stay
Java: files that *declare* public API customers and external kit authors compile against,
plus `package-info.java`, which has no Kotlin equivalent. Those lines are measured
separately, so "Java left to migrate" can honestly reach zero while the facades stay put.

Adding a file to this list is an API-surface decision and needs an SDK owner's review.

**`kotlin-migration-progress` CI job** — runs on every PR. It measures the PR head *and*
the base commit, then upserts a single sticky PR comment with:

- how many lines this PR moved from Java to Kotlin
- the per-module table and the overall goal percentage
- the 15 largest Java files still outstanding

The base is always measured with the PR's *current* facade list, so editing that list can
never masquerade as migration progress. The same content goes to the job summary.

**Ratchet** — the job fails if a PR increases Java LOC in these three modules. New code
here should be Kotlin. Escape hatch: the `allow-new-java` label plus a reason in the
description.

## Baseline it reports today

| Module | Kotlin LOC | Java LOC | Java facade (staying) | Java left to migrate | Kotlin % |
| --- | ---: | ---: | ---: | ---: | ---: |
| `android-core` | 3,225 | 22,457 | 9,241 | **13,216** | 12.6% |
| `android-kit-base` | 557 | 5,380 | 1,651 | **3,729** | 9.4% |
| `kits/rokt/rokt` | 1,122 | 0 | 0 | **0** | 100.0% |
| **Total** | **4,904** | **27,837** | **10,892** | **16,945** | **15.0%** |

Migration goal progress: **22.4%**.

## Before merging

- [ ] Create the `allow-new-java` label — without it the ratchet override has no handle.
- [ ] Confirm the base branch. This targets `workstation/kotlin-migration`, the long-lived
      integration branch. If you'd rather have the tracker commenting on *every* PR
      immediately, retarget to `main` — it's a one-click base change and the job is
      additive.
- [ ] Sanity-check the facade list. It is my read of the public surface, not gospel:
      53 core files and 6 kit-base files. `KitIntegration`, `KitUtils`, `ReportingMessage`,
      `CommerceEventUtils`, `FilteredMParticleUser` and `FilteredIdentityApiRequest` are on
      it because 50, 26, 28, 19, 4 and 16 external kit files reference them respectively.

## Notes

- Job-level `permissions: pull-requests: write`; the workflow default stays `read`.
- Comment upsert uses `gh api` with `GITHUB_TOKEN` — no new third-party action.
- Comment step is skipped on fork PRs, where the token is read-only. The job summary
  still renders.
- Script is shellcheck-clean and was verified locally in both `--format` modes and in
  `--root` mode against a detached worktree, which is how the job measures the base.
- `docs/kotlin-migration.md` is included here because it defines the facade list's
  rationale. Happy to split it into its own PR if you'd prefer this one be pure CI.

### DIFF
--- a/docs/kotlin-migration.md
+++ b/docs/kotlin-migration.md
@@ -1,0 +1,423 @@
+# mParticle Android SDK — Java → Kotlin migration
+
+**Status:** planning · **Owner:** Android SDK team · **Started:** Sep 2, 2026 · **Target:** Oct 9, 2026
+
+This is the living tracker. Strike through items and append the PR link as they land,
+the same way the Apple SDK's Swift migration doc works.
+
+---
+
+## 1. Goal
+
+Migrate `android-core`, `android-kit-base`, and the Rokt kit to Kotlin **without changing
+the public interface**. Every line that is not part of the customer- or kit-author-facing
+API surface should be Kotlin by the end of the migration.
+
+This is the Android translation of the principle the Apple SDK is following on
+`workstation`: *do not attempt a wholesale rewrite — slice pure subsystems out and leave
+the public facade in the original language as the orchestrator.*
+
+For Apple, ObjC keeps the singleton, `sqlite3`, and dynamic dispatch. For Android,
+**Java keeps the public API declarations** — the classes and signatures that customers and
+external kit authors compile against — and everything behind them becomes Kotlin.
+
+### Baseline (Sep 2, 2026)
+
+| Module | Kotlin LOC | Java LOC | Java that stays (public facade) | Java left to migrate | Kotlin % |
+| --- | ---: | ---: | ---: | ---: | ---: |
+| `android-core` | 3,225 | 22,457 | 9,241 | **13,216** | 12.6% |
+| `android-kit-base` | 557 | 5,380 | 1,651 | **3,729** | 9.4% |
+| `kits/rokt/rokt` | 1,122 | 0 | 0 | **0** | 100.0% |
+| **Total** | **4,904** | **27,837** | **10,892** | **16,945** | **15.0%** |
+
+**Migration goal progress: 22.4%** — Kotlin as a share of everything that is not a
+designated public-API facade. This is the number the CI job reports, and it is the number
+that needs to reach 100%.
+
+> The Rokt kit is already 100% Kotlin (8 files, 1,122 LOC in `src/main`). Nothing to do
+> there beyond keeping it that way, which the CI ratchet now enforces.
+
+### Out of scope for the short-term migration
+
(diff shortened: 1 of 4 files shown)


thomson-t (approved):

docs/kotlin-migration.md
```diff
@@ -0,0 +1,445 @@
...
+1. **One file per PR.** Stack them with Graphite inside your lane. Do not touch a file
+   that belongs to another lane; if you need a change there, ask the lane owner to stack
+   it, or note it in the Cleanup section.
+2. **Characterization test first.** If the file has no meaningful test coverage, the PR
+   _below_ it in the stack adds tests against the current Java behaviour. Converting an
+   untested file is how a silent behaviour change ships.
+3. **Convert, do not refactor.** Use IntelliJ's _Convert Java File to Kotlin File_, then
+   clean up only what the compiler and ktlint demand. Renames, redesigns, coroutine
+   adoption, and null-safety improvements are all separate follow-up PRs.
+4. **Land the file in `src/main/kotlin/`,** matching the package path. (We currently have
```
jamesnrokt, line 100: nit: I'd suggest we remove the separation entirely here. There's no real need to have kotlin and java separated

jamesnrokt (approved):

docs/kotlin-migration.md
nickolas-dimitrakas: Agreed, and done in #771 (0592e1e6) since the doc moved there.

The rule is now "convert in place" — leave the file where it is and change the extension. Both `src/main/java` and `src/main/kotlin` are already configured as Kotlin source directories in every module, and Kotlin compiles from both today, so there was never a separation worth maintaining.

Bonus reason it is the better rule: it keeps a rename out of every conversion diff. A rename plus a content change is exactly the shape that hides a signature change from a reviewer, and we are about to do ~130 of these. The cleanup item to relocate the five existing stragglers is gone too — there is nothing to clean up.

PR #773, opened 2026-09-02
docs: timebox the migration to two weeks

nickolas-dimitrakas:
Stacked on #771. Compresses the schedule from six weeks to **Wed Sep 2 → Wed Sep 16**.

That is **ten working days, not eleven — Mon Sep 7 is Labor Day.** 16,945 lines at ~1,700
a day, with no recovery week. Moving the dates alone would not have made that arithmetic
work, so four things change with them.

## What changes

**1. Batch the easy tier per lane directory.** 41 leaf files — table schemas, message
types, small interfaces — become ~8 PRs instead of 41. This is the single biggest lever,
and it is only safe because #770's ratchet blocks new Java and #771's API guard reports
exactly which signatures moved. A batch whose guard output is anything but "no diff" gets
split before review. Medium and hard files stay one per PR.

**2. Lane H runs in parallel, not last.** A two-week box cannot serialise 3,412 lines
behind everything else, and it does not have to: lanes are file-scoped and Java and Kotlin
interoperate inside one module, so nobody is blocked on H. The "goes last" rule was
conservatism. Residual risk is a same-day semantic conflict between a caller in another
lane and an H callee — tests catch that, the guard does not — so H's owner announces each
merge.

**3. Staffing is now explicit.** Five engineers at ~340 lines each per day:

| Engineer | Lanes | LOC |
| --- | --- | ---: |
| 1 | F — kit base internals | 3,729 |
| 2 | H — cross-cutting | 3,412 |
| 3 | D + C — event pipeline, identity | 3,633 |
| 4 | A + E — persistence, kit infra core side | 3,490 |
| 5 | B + G — networking, platform surface | 2,681 |

With four, that is ~424 lines per engineer per day and no slack — drop lane B or G rather
than pretending. **With three, two weeks is not achievable** and the honest move is to keep
the six-week plan.

**4. Review turnaround becomes a gate item.** At ~60–80 PRs in ten days, throughput — not
conversion speed — is the binding constraint. Each lane gets a named reviewer with a
four-hour turnaround, and mechanical conversions are approved on **evidence** (green API
guard, green tests, the diff is the same code in the other language) rather than a
line-by-line reading of a language port.

## Milestones

| | Date | Work | Goal |
| --- | --- | --- | ---: |
| M0 | Thu Sep 3 | Setup; **every owner has one PR merged** | 22.4% |
| M1 | Fri Sep 4 | Easy tier, every lane · 2,581 | 34.3% |
| M2 | Wed Sep 9 | Medium tier, every lane · 5,758 · **checkpoint** | 60.6% |
| M3 | Fri Sep 11 | Hard tier first half · 3,753 · smoke QA | 77.8% |
| M4 | Tue Sep 15 | Hard tier second half · 4,853 · **internal Java = 0** | 100% |
| M5 | Wed Sep 16 | Full QA, whole-migration guard check, release | — |

M0 deliberately requires every owner to merge one conversion PR on day one. The point is
to find the friction in review and CI immediately, not on day six.

## Pressure valve

If the tracker is short of 60% at the **Sep 9** checkpoint, cut in this order and say so in
the channel the same day — decided then, not on Sep 15:

1. `KitConfiguration` + `KitManagerImpl` (2,510) — biggest and riskiest, referenced by 29
   and 12 kit files. Costs 15 points of goal metric, keeps the release safe.
2. `MParticleJSInterface` (840) — highest blast radius per line; the WebView bridge
   resolves its methods by name from JS outside this repo.
3. Lane G medium (1,236) — lowest value per line, most platform-coupled.

Cutting all three still lands ~72%, with every easy and medium file converted and the
public API provably unchanged. That is a good two-week outcome. Missing the QA pass
because the last hard file landed on Sep 16 is not.

## Explicitly outside the timebox

Deferred on purpose, not forgotten — none of it is required for internal Java to reach
zero:

- **Lane I, facade thinning** (~5,000 LOC). The most delicate work in the project. It
  moves the Kotlin % and not the goal metric, so it can follow the release.
- `testutils` (43 files) and the three remaining Java files in test source sets.
- Cleanup items.

## QA

Three passes rather than two, with the new one where it earns its keep:

- **M2, Wed Sep 9 — full pass.** Straight after the medium tier. That tier is where the
  behaviour-carrying code lives — network clients, identity, DB services, push, device
  attributes, kit mappings — and 5,758 lines of it land in five days. A regression found
  on Sep 9 has a week of runway; the same one found at the Sep 16 release gate does not.
  A QA failure at M2 counts as being behind, and triggers the pressure valve.
- **M3, Fri Sep 11 — smoke pass.** First four items, half an hour, on the integration
  branch, after the first hard-tier files.
- **M5, Wed Sep 16 — full pass.** The release gate.

### DIFF
--- a/docs/kotlin-migration/PLAN.md
+++ b/docs/kotlin-migration/PLAN.md
@@ -1,6 +1,6 @@
 # mParticle Android SDK — Java → Kotlin migration
 
-**Status:** planning · **Owner:** Android SDK team · **Started:** Sep 2, 2026 · **Target:** Oct 9, 2026
+**Status:** planning · **Owner:** Android SDK team · **Timebox:** Wed Sep 2 → Wed Sep 16, 2026 (10 working days)
 
 This is the living tracker. Strike through items and append the PR link as they land,
 the same way the Apple SDK's Swift migration doc works.
@@ -88,9 +88,17 @@ the Java file thins out to declarations that delegate. They are just excluded fr
 
 ## 3. How to do a conversion
 
-1. **One file per PR.** Stack them with Graphite inside your lane. Do not touch a file
-   that belongs to another lane; if you need a change there, ask the lane owner to stack
-   it, or note it in the Cleanup section.
+1. **Batch the easy tier, one file per PR for everything else.** Easy-tier files are
+   leaf classes — table schemas, message types, small interfaces — and reviewing seven of
+   them together is no harder than reviewing one. Batch them per lane directory: roughly
+   8 PRs instead of 41. Medium and hard files stay one per PR, stacked with Graphite
+   inside your lane. Never touch a file that belongs to another lane; ask that owner to
+   stack it, or note it in Cleanup.
+
+    This is the change that makes a two-week timebox arithmetic rather than hope. It is
+    only safe because the API guard reports exactly which signatures moved and the ratchet
+    blocks new Java — without both, batch conversions are how a silent break ships.
+
 2. **Characterization test first.** If the file has no meaningful test coverage, the PR
    _below_ it in the stack adds tests against the current Java behaviour. Converting an
    untested file is how a silent behaviour change ships.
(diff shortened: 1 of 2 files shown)


nickolas-dimitrakas:
Folding this into #771 instead — no separate PR, and the dated milestones are coming out.

With three devs and priorities that move, a calendar is a liability: the plan reads as failed the moment something else takes precedence, even though the work is fine. Milestones become gates on the percentage the CI tracker prints, so the plan can be put down and picked back up at whatever number it was left at.

Branch kept in case anything needs recovering from it.

PR #812, opened 2026-09-11
perf: allow apps to shrink unused SDK code

thomson-t:
## Background

Package-wide consumer keep rules prevent integrating apps from removing unused SDK code. Narrow the rules to the runtime entry points that need protection while preserving the existing stack-trace-based diagnostic implementation.

## What Has Changed

- Replace core's blanket keep and warning suppression with targeted rules for JavaScript bridges, Parcelable creation, optional integrations, kit discovery, and diagnostic stack names and frames.
- Package kit-base consumer rules and include compatibility coverage in core for older kit-base artifacts. Export consumer rules in every core and kit-base variant.
- Preserve the identity helper called by kit-base across the producer-obfuscated core AAR boundary.
- Remove the Rokt kit's package-wide keeps and update the shrinking documentation.
- Align Rokt SDK and payment extension at `6.1.2`, including the kit and SDKPlus standalone defaults.

## Screenshots/Video

N/A — no UI changes.

## Checklist

- [x] Self-review completed
- [ ] Tests added or updated
- [x] Tested locally

## Additional Notes

Validation:

- `./gradlew test lint ktlintCheck :android-core:assembleRelease :android-kit-base:assembleRelease` passed. Core and kit-base passed 310 and 89 unit tests respectively in each debug/release variant; tooling and plugins passed four more tests.
- Rokt kit release build and 92 release unit tests passed against locally published mParticle artifacts and the published Rokt SDK `6.1.2` dependency.
- SDKPlus build and local publication passed. Generated POMs and resolved dependencies confirm Rokt SDK and payment extension `6.1.2`, with UX Helper `2.0.2`.
- Minified kit and SDKPlus fixture builds passed. Against the same fixtures with both Rokt artifacts pinned to `6.1.1`, their APKs shrank by 131,718 and 237,550 bytes respectively. R8 emitted Kotlin metadata warnings with both dependency versions.
- Inspected packaged consumer rules in all three AARs, core's producer mapping, and the exported identity helper signature. Confirmed the Java/Kotlin source trees are unchanged.
- `trunk check --ci` passed with no new findings (21 existing findings). Android lint emitted Kotlin metadata-version diagnostics despite completing successfully; Gradle/dependency deprecation warnings remain.

Core and kit-base should be released together. Instrumented tests, minified runtime rendering verification, and licensed DexGuard validation are not part of this verification. Local harness and fixture changes are excluded from the PR.

### DIFF
--- a/android-core/consumer-proguard.pro
+++ b/android-core/consumer-proguard.pro
@@ -48,7 +48,7 @@
 
 # InternalListenerManager identifies SDK frames by package and reads API names
 # from the live stack. Preserve those names/frames without retaining unused APIs.
--keeppackagenames com.mparticle.**
+-keeppackagenames com.mparticle,com.mparticle.**
 -keep,allowshrinking class com.mparticle.MParticle {
     public <methods>;
 }
--- a/android-core/consumer-proguard.pro
+++ b/android-core/consumer-proguard.pro
@@ -1,3 +1,96 @@
-#these are the rules that are packaged with the Core SDK .aar to be used in consuming apps.
--keep class com.mparticle.** { *; }
--dontwarn com.mparticle.**
+# SDK-owned runtime entry points. Ordinary JSON models use explicit org.json keys.
+-keepattributes RuntimeVisibleAnnotations
+-keepclassmembers,allowoptimization class com.mparticle.internal.MParticleJSInterface {
+    @android.webkit.JavascriptInterface <methods>;
+}
+-keepclassmembers,allowoptimization class com.mparticle.messaging.ProviderCloudMessage {
+    public static final android.os.Parcelable$Creator CREATOR;
+}
+
+# Optional advertising-ID and legacy Firebase methods invoked reflectively.
+-keep,allowoptimization class com.google.android.gms.ads.identifier.AdvertisingIdClient {
+    public static com.google.android.gms.ads.identifier.AdvertisingIdClient$Info getAdvertisingIdInfo(android.content.Context);
+}
+-keep,allowoptimization class com.google.android.gms.ads.identifier.AdvertisingIdClient$Info {
+    public java.lang.String getId();
+    public boolean isLimitAdTrackingEnabled();
+}
+-keep,allowoptimization class com.google.firebase.iid.FirebaseInstanceId {
+    public static com.google.firebase.iid.FirebaseInstanceId getInstance();
+    public java.lang.String getToken(java.lang.String, java.lang.String);
+}
+
+# Exact presence probes; method calls on these optional libraries are direct.
+-keep,allowoptimization class com.google.firebase.messaging.FirebaseMessaging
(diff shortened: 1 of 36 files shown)


android-core/consumer-proguard.pro
```diff
@@ -1,3 +1,96 @@
...
+
+# These compileOnly APIs are guarded by the presence checks above.
+-dontwarn com.android.installreferrer.api.InstallReferrerClient
+-dontwarn com.android.installreferrer.api.InstallReferrerClient$Builder
+-dontwarn com.android.installreferrer.api.InstallReferrerStateListener
+-dontwarn com.google.android.gms.tasks.OnFailureListener
+-dontwarn com.google.android.gms.tasks.OnSuccessListener
+-dontwarn com.google.android.gms.tasks.Task
+-dontwarn com.google.android.instantapps.InstantApps
+-dontwarn com.google.firebase.messaging.FirebaseMessaging
```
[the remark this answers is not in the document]
thomson-t: These classes are still needed.
Previously, installreferrer was optional too. The source and runtime availability checks are unchanged. However, the old -dontwarn com.mparticle.** suppressed missing-class diagnostics originating from SDK classes—even when the missing type belonged to another package. Removing that blanket suppression changes the warning coverage.

android-core/consumer-proguard.pro
```diff
@@ -1,3 +1,96 @@
...
+}
+# A Class token registers custom kits, so unused implementations can still shrink.
+-keep,allowobfuscation,allowshrinking class * extends com.mparticle.kits.KitIntegration
+-keepclassmembers class * extends com.mparticle.kits.KitIntegration {
+    public <init>();
+}
+
+# InternalListenerManager identifies SDK frames by package and reads API names
+# from the live stack. Preserve those names/frames without retaining unused APIs.
+-keeppackagenames com.mparticle.**
```
BrandonStalnaker: This misses the com.mparticle package itself.

ProGuard/R8 package filters are literal-prefix plus wildcard, so com.mparticle.** requires the trailing dot and does not match the package com.mparticle.

Not a huge deal since InternalListenerManager.start() only runs when MPUtility.isAppDebuggable(context) or the debug.mparticle.listener prop matches, so this is diagnostics-only, not a release-runtime bug.

BrandonStalnaker (commented):
just one thing I was unsure about

android-core/consumer-proguard.pro
thomson-t: Thanks. I will make it explicit.
Although in my test with R8 version 8.9.35 in full mode, it still kept the root package.
But better to make it separate like this.
`-keeppackagenames com.mparticle,com.mparticle.**`

nickolas-dimitrakas (dismissed):
LGTM

BrandonStalnaker (dismissed):

nickolas-dimitrakas (approved):

PR #817, opened 2026-09-17
fix(lint): restrict data-plan detector to known event builder types

nickolas-dimitrakas:
## Background
The Data Plan lint detector evaluates `MPEvent.Builder`/`CommerceEvent.Builder` calls found in analyzed source to check them against a data plan. This change tightens what it's allowed to touch while doing that evaluation.

## What Has Changed
- Restrict the detector's expression evaluation to an explicit allowlist of mParticle event/commerce builder types.
- Add a bounded recursion depth to expression resolution.

## Screenshots/Video
N/A — lint tooling change, no UI.

## Checklist
- [x] Self-review completed
- [x] Tests added
- [x] Tested locally

## Additional Notes
- `./gradlew :tooling:custom-lint-rules:test` and `:tooling:custom-lint-rules:ktlintCheck` pass.
- New regression tests in `DataplanDetectorTest` confirm the allowlist is enforced and that a normal builder chain still resolves as before.


tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/MethodCall.kt
```diff
@@ -10,10 +10,10 @@ data class MethodCall(override val parent: Expression, val methodName: String?,
         arguments.forEach { it.parent = this }
     }
 
-    override fun resolve(): Any? {
+    override fun resolve(): Any? = ResolutionGuard.guarded {
```
thomson-t, line 13: MethodCall.resolve() is never gated, and that’s the widest call site.

resolves its parent and then invokes any method matching name + arity on whatever object comes back, via instance::class.java.methods — which includes inherited Object methods and statics. So the allowlist can be walked around without ever defeating it:
`new MPEvent.Builder("x").getClass().forName("some.Class.On.Lint.Classpath").newInstance()`

That’s valid Java (static-through-instance is legal; Class.newInstance() is a real instance method), it passes CallScanner.ofInterest, and resolveChainedCalls folds it into Constructor → MethodCall(getClass) → MethodCall(forName) → MethodCall(newInstance). getClass() returns a java.lang.Class, and from that point the receiver type is Class, so forName / getResource / newInstance all match on name and arity. Result: arbitrary static-initializer execution plus no-arg instantiation of anything on the lint classpath, and each subsequent chained call then invokes name/arity-matched methods on that new object with literal arguments.

Suggested fix: reject in MethodCall.resolve() unless instance::class.java.name is allowlisted, before the method search. That’s one guard at the shared chokepoint rather than per-call-site. AllowedTypes needs the value types that allowlisted methods legitimately return (MPEvent, CommerceEvent, String, boxed primitives, Double for Product.getTotalAmount(), the allowed collections).

tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/Constructor.kt
```diff
@@ -7,9 +7,12 @@ import org.jetbrains.uast.UCallExpression
 data class Constructor(override val parent: Expression, val methodName: String?, override val node: UCallExpression) : ParameterizedExpression {
     override var arguments: List<Value> = listOf()
 
-    override fun resolve(): Any? {
+    override fun resolve(): Any? = ResolutionGuard.guarded {
```
thomson-t, line 10: Still returns clazz on instantiation failure.

A second route to a Class instance that doesn’t even need getClass(): new MPEvent.Builder(0) → getConstructor(Integer) misses → arity-1 fallback picks Builder(String) → newInstance(Integer) throws → caught → the Class is returned. Same escalation as above. Return null instead.

tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/StaticFactory.kt
```diff
@@ -9,8 +9,11 @@ class StaticFactory(val methodName: String?, override val node: UCallExpression)
     override val parent = RootParent(node)
     override var arguments: List<Value> = listOf()
 
-    override fun resolve(): Any? {
+    override fun resolve(): Any? = ResolutionGuard.guarded {
```
thomson-t, line 12: StaticFactory’s Kotlin allowlist entries look functionally wrong

StaticFactory derives the name from (node.resolve()?.parent as? ClsClassImpl)?.stub?.qualifiedName, which for a Kotlin multifile facade gives the part class — kotlin.collections.MapsKt__MapsKt, not MapsKt. (Extensions.kt:176 has to split("__")[0] for exactly this reason.) If that holds, mapOf(...) / listOf(...) attribute maps now resolve to null and feed straight into item 3 — the existing testCollection won’t catch it because it never builds an event. There’s no Kotlin test in the PR either way, so worth confirming. Separately, MapsKt / CollectionsKt are allowed at class granularity with declaredMethods + isAccessible = true; allowlisting method names rather than classes would fix both the breadth and the facade-name problem.

thomson-t (dismissed):

tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/MethodCall.kt
nickolas-dimitrakas: Good catch. Gated `MethodCall.resolve()` on `AllowedTypes.isInstanceAllowed(instance)` before searching for a matching method, so a call onto anything not in the allowlist (e.g. the `Class` returned by `getClass()`) is rejected before it's invoked. Added `testReflectionEscapeViaGetClassIsNeverExecuted`, which reproduces this exact chain and asserts `Class.forName` never runs (verified it fails without the fix).

tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/Constructor.kt
nickolas-dimitrakas: Fixed - both the "no matching constructor" and "newInstance threw" paths now return null instead of the raw `clazz`. Covered by the same instance-gating design as the getClass() fix: nothing should ever hand back a live Class object.

tooling/custom-lint-rules/src/main/java/com/mparticle/lints/dtos/StaticFactory.kt
nickolas-dimitrakas: Confirmed by directly inspecting the compiled stdlib: `kotlin.collections.MapsKt` declares zero methods of its own (it just extends the part-class chain), and `mapOf` actually lives on `MapsKt__MapsKt`. `StaticFactory` now normalizes the "__Part" suffix for the allowlist check (still uses the original name for `Class.forName`), and checks a specific method-name allowlist per facade instead of every declared method. Added `AllowedTypesTest` covering the real part-class name, the facade-name normalization, and rejection of both unrelated classes and non-allowlisted methods on an allowed class.

thomson-t (approved):

